PR #58, opened 2014-04-29
GCW-Zero support

Shin-NiL:
I got working binaries for [GCW-Zero](http://www.gcw-zero.com/). It seems ok to me.
Is this the correct approach to add a new toolchain? If it's not, please ignore this request.


Shin-NiL:
@jgranick Any help here? 

What should be my next step to port OpenFL? Aether? Still no clear to me :(

PR #77, opened 2014-06-21
Update hxString.h

porfirioribeiro:
Added cast's to `unsigned char*` and `char*`
When using @:native on some functions that require this types.

### DIFF
--- a/include/hxString.h
+++ b/include/hxString.h
@@ -93,6 +93,8 @@ public:
    const char *__CStr() const;
    const wchar_t *__WCStr() const;
    inline operator const char *() { return __s; }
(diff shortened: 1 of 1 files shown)


hughsando:
You can now type your externs with "cpp.CastCharStar", which will accept a String, like in this example:
```
extern class Extern
{
   @:native("strcat")
   public static  function strcat( base:cpp.CastCharStar, add:cpp.ConstCharStar ) : Void;
}

class Test
{
   public static function main()
   {
      Extern.strcat("Hello", "goodbye");
   }
}
```
This code will compile - but it will crash at runtime because changing a hxcpp string is really really dangerous.  But I'm assuming the API is just not fully annotated with "const".

PR #109, opened 2014-09-24
alloc_null() test

codeservice:
Have new interesting TestNull.hx for support. Works on cpp, fail on
neko.


hughsando:
have included this test in test/cffi.
I think the original problem is the lack of call to neko_init.  This should also print a better message now.

PR #122, opened 2014-10-23
Changed name of enum ValueType

nensanders:
It was conflicting with generated cpp class when using Reflect.* together with native extensions.
screen shot 2014-10-23 at 16 48 11


hughsando:
I have renamed it - "ValueType" does seem a bit common.  I don't think I can put it in a namespace because code may be referring to the values in the global namespace.  Thanks for bringing it to my attention.

PR #128, opened 2014-11-17
Make StdLibs.cpp compile under MinGW

Draknek:
Using the MinGW compiler, the macro `_WIN32_WINNT` needs to be defined as 0x0501 or higher to get the `AllocConsole` function. Without this, it just won't compile at all.


hughsando:
I now use dynamic linking to get this to work with mingw.  Does this solve the issue?

hughsando:
I'm closing this one - feel free to reopen it if the problem persists.

PR #138, opened 2014-12-03
Fix: add a basic mapping from haxe -> c++

fzzr-:
e.g.: `int = Int`.
When we build library (with `-D static_link`),
It allows to use it in Xcode-project and safely build.

_Mb. need more types to add._


fzzr-:
screen shot 2015-01-09 at 18 27 26

hughsando:
I have moved the hxcpp 'Class' to 'hx::Class', so you can compile with objc now.  Does this solve this issue?

hughsando:
I think objc compiling should be good now.  Perhaps start a new issue if there are other concerns.

fzzr-:
Thx. I'll test it soon.

PR #151, opened 2014-12-31
explicit namespace references to avoid naming conflicts with iOS Objective C

cbatson:
This allows hxcpp.h to be included in a .mm file.


cbatson:
I've been unsuccessful at finding an elegant way to deal with this. I will just keep this change local for now.

For reference, the `typedef` of `Class` in the global namespace (in hxcpp.h) conflicts with a definition in one of Apple's headers.  Ideally, hxcpp would not put definitions in the global namespace.  (And neither would Apple!)

PR #160, opened 2015-01-17
allow setting of HXCPP_DEBUG_INFO to generate debug info for iOS

cbatson:
This is useful, for example, in creating release builds with symbols so that crash dumps can be symbolicated.


hughsando:
I have reused the HXCPP_DEBUG_LINK variable for this, which is the same one msvc uses.  This will generate a -g and a -O2, I guess this is what you want.

PR #185, opened 2015-03-14
Mac tests & issue 65

thomasuster:


hughsando:
The array sort issue is a bit complicated because the user-supplied compare function may do arbitrary work, including may GC operations, so it is not really feasible to disable the GC in this case.

thomasuster:
GC operations like GC.run?

PR #196, opened 2015-03-25
msvc: fix XP subsystem flags, add x64 XP correction

underscorediscovery:
This commit makes three important fixes for WinXP support;

1) rename `CONSOLE_VER` -> `SUBSYSTEM_VER` (it's not specific to
console subsystem)
1) `-subsystem` ${ver} value is **required** for both types for XP
compat
2) Adds the x64 variation for windows XP x64 builds (5.02, x86 is 5.01)

MSDN has all the details to verify,
I tested a plain 32 bit build (built with vs2013 on Windows 8.1),
Running on XP 32 bit: It fails to run with "not a valid win32
application".

Adding the correct subsystem flags makes it work as expected under the
same setup.
More info here if needed:
https://msdn.microsoft.com/en-us/library/fcc1zstk.aspx

### DIFF
--- a/toolchain/msvc-toolchain.xml
+++ b/toolchain/msvc-toolchain.xml
@@ -24,8 +24,12 @@
 <set name="SUBSYSTEMWINDOWS" value="1" if="no_console" unless="HXCPP_DEBUGGER" />
 <set name="SUBSYSTEMCONSOLE" value="1" unless="SUBSYSTEMWINDOWS" />
 
-<set name="CONSOLE_VER" value=",5.01" if="HXCPP_WINXP_COMPAT" />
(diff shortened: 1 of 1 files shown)


hughsando:
I have merged this in, and allowed XP_COMPAT for x64.  Now, I default to xp compat for 32 bit, but you need to explicitly put it on for 64 bit.  Not sure how many people are making 64 bit exes for XP, but there is at least an option now.  I generated the files and they ran on win 8.1.  Have not tested them on a winxp box though.
Thanks.

underscorediscovery:
Sounds good (the 64bit fix was just a side effect since the value differs)

PR #197, opened 2015-03-26
Topic add getbytes

bjitivo:
This function is used by haxe std library changes that we will be creating a separate pull request for with the HaxeFoundation.  It is intended to make some array operations faster/less churny.

### DIFF
--- a/include/Array.h
+++ b/include/Array.h
@@ -471,6 +471,15 @@ public:
       }
    }
 
+   void addBytes(const Array<ELEM_> &inArray, int pos, int len)
+   {
+       int orig_len = length;
+       this->EnsureSize(orig_len + len);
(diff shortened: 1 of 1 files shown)


hughsando:
This function is covered by the array "blit" function, which is used by the byte array.  The blit function takes an extra "destination position" parameter:
```
using cpp.NativeArray;
class Test
{
   public static function main()
   {
      var a = [1,2,3,5,6];
      var b = [7,8,9,10];

      var pos = 1;
      var len = 3;

      //a.addBytes( b, pos, len );
      a.blit(a.length, b, pos, len );
      trace(a); // [1,2,3,5,6,8,9,10]
   }
}
```
The "using cpp.NativeArray" syntactically exports the blit function to an array member.

bjitivo:
Good point.  Not sure how I missed that.  Thanks.

PR #199, opened 2015-03-26
Topic enum constructor throw

bjitivo:
Causes code which creates enums at runtime (using reflection I think, it's been a while since this change was made, and I can't remember the details) to throw a runtime exception that is more useful than the one that it was throwing previously.  Also breaks into the debugger if it's running.

### DIFF
--- a/include/hx/Debug.h
+++ b/include/hx/Debug.h
@@ -545,7 +545,11 @@ inline void __hxcpp_dbg_setAddStackFrameToThreadInfoFunction(Dynamic) { }
 // created and terminated
 inline void __hxcpp_dbg_threadCreatedOrTerminated(int, bool) { }
 
-Dynamic __hxcpp_dbg_checkedThrow(Dynamic toThrow);
+inline Dynamic __hxcpp_dbg_checkedThrow(Dynamic toThrow)
+{
+    hx::Throw(toThrow);
+    return null();
+}
 
 #endif // HXCPP_DEBUGGER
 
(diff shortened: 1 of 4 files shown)


bjitivo:
Any comments on this?  It's nicer to get a real exception with a useful message instead of what you get prior to this change, I forget what exactly, but something nearly impossible to debug since it's a very general error that doesn't indicate that the exception was a bad enum construction.

hughsando:
I hand merged this one - but it looks good.

bjitivo:
Just curious, why hand merging?  Is it because my pull requests are weird?  I am still trying to get the hang of how git manages its topic branches and doing rebases and stuff.

jgranick:
Another way to do it (avoiding the "Merge pull request" messages) would be to cherry-pick:

    git remote add tivo https://github.com/tivo/hxcpp
    git fetch tivo
    git cherry-pick c26b606b51d5fa1dfb63ba3bb2610b1dce50ece2
    git cherry-pick 517ba61789560565bb63df3a21c91dbb4f8b9542

hughsando:
Yes, it was the commit list that spooked me.  If I could be sure it was just the 'Files Changed' that would change I would gladly have them magically appear in the repo, but I've had bad experiences with git.
I started to inspect the change list and saw this with no body:
+ Dynamic __hxcpp_dbg_checkedThrow(Dynamic toThrow);
And I thought that something was wrong (corrected with second commit) - I see I should have looked at the aggregate "files changed" list.  For me, It only took a minute to merge the changes, but I could see hours of frustration if the commit somehow reverted some unrelated file to an earlier version.
I probably should have downloaded a patch and applied it locally - then I could see exactly what the state would be after the changes.

thomasuster:
I ask people to "squash" their commits before looking at them.

git reset head~4 --soft
git commit -am "Topic enum constructor throw"

bjitivo:
Thank you, this is very helpful.  I am sure that my pull requests will improve with this feedback!

PR #200, opened 2015-03-26
Topic fix debugger breakpoint hash

bjitivo:
We found that the HXCPP_DEBUG_HASHES functionality wasn't working.  I would expect that if this is the case, you'd know about it already as others would have reported it.  So it's possible that this change is unnecessary.  At the very least, this change makes the code compile and work correctly when HXCPP_DEBUG_HASHES is not defined.  You can take out the part of the change that disables HXCPP_DEBUG_HASHES (which is the first diff listed below, that comments out #define HAXPP_DEBUG_HASHES) if you think that debug hashes really are working, but the rest of the change should be taken regardless.

### DIFF
--- a/include/hx/Debug.h
+++ b/include/hx/Debug.h
@@ -122,6 +122,9 @@ public:
          #ifdef HXCPP_DEBUG_HASHES
          classFuncHash(inClassFunctionHash),
          fileHash(inFileHash),
+         #else
+         classFuncHash(0),
+         fileHash(0),
          #endif
          fullName(inFullName), fileName(inFileName),
          #ifdef HXCPP_STACK_LINE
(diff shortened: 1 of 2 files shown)


bjitivo:
Cleaned up pull request, rebased etc.

bjitivo:
I tested and found that breakpoints *are* working, so I think that the original bug was fixed somewhere along the way.  However, this change is still worth taking because it ensures that if HXCPP_DEBUG_HASHES is not defined for some reason, the debugger still works.  Without this change, not having that symbol defined results in a nonfunctioning debugger.

hughsando:
Might give this a year or so, and then assume HXCPP_DEBUG_HASHES will always be defined and remove the conditionals.

PR #201, opened 2015-03-26
Topic gc collect compact

bjitivo:
We need these functions in our own application, so exposing them.

### DIFF
--- a/include/hx/CFFIAPI.h
+++ b/include/hx/CFFIAPI.h
@@ -134,6 +134,8 @@ DEFFUNC_0(void, gc_enter_blocking)
 DEFFUNC_0(void, gc_exit_blocking)
 DEFFUNC_0(void, gc_safe_point)
 DEFFUNC_2(void, gc_set_top_of_stack,int *,bool)
+DEFFUNC_0(void, gc_collect)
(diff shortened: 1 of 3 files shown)


hughsando:
Not sure about this one.  Since it does not rely on any special types, you can export any function to like from hxcpp back into native code by passing in a closure.

underscorediscovery:
I'd be interested in more fine grained access to gc (Is gc disable is still a nop?) I've not yet gotten to this but I know for sure in future I need to turn off GC at times and I think that access to collect/compact compliments this need as well. 

Do you mean that calling within a closure would have access to the gc calls with untyped __cpp__ ?

hughsando:
I was thinking more about how to export the function to the ndll code, so
from cpp:

myndll_set_gc_runner( function() cpp.vm.Gc.run(true) )
myndll_set_gc_enabler( cpp.vm.Gc.enable )

And then store this closure in the ndll somewhere.

underscorediscovery:
Ah ok, but in my previous tests only run works, does enable work yet? I would use that if I could

bjitivo:
OK, we'll try to find another way to do this.  Feel free to close this pull request.

PR #203, opened 2015-03-26
Topic hxcpp api level

bjitivo:
If HXCPP_API_LEVEL is not defined at all, this code would fail to compile. I don't recall when this would happen, but regardless, the change here is simple and guards against that.


hughsando:
I have ensured that HXCPP_API_LEVEL is always defined now.

PR #204, opened 2015-03-26
Topic hxcpp visit allocs

bjitivo:
This change makes the code that is emitted when HXCPP_VISIT_ALLOCS is not defined (which we do not define, because we want the binary to be as small as possible, and we don't use the "visit allocs" functionality) smaller, by conditionally dis-including more unneeded functions.


Simn:
CI fails:

```
Error: ./src/unit/issues/Issue4002.cpp: In member function ‘virtual void unit::issues::Issue4002_obj::fail(Dynamic)::_Function_1_1::__Visit(hx::VisitContext*)’:
./src/unit/issues/Issue4002.cpp:64:2: error: ‘v0’ was not declared in this scope
Error: Build failed
```

bjitivo:
The builds are passing now.  Any comments?  I know that this introduces a bunch of ugly #ifdefs but we are concerned about code size and do not want the visit functions to be emitted all over the place when they are not necessary.

bjitivo:
Rebased, cleaned up pull request.

PR #205, opened 2015-03-26
Topic macros 27

bjitivo:
For some reason, I can't remember why, we needed these macros to take up to 27 arguments instead of 21.


bjitivo:
Any comments?

bjitivo:
Rebased, cleaned up pull request.

PR #206, opened 2015-03-26
Topic no pcre debug

bjitivo:
When we define DEBUG, which we do for our own purposes, the PCRE library was emitting tons of debugging stuff.  So make the pcre library's debug output not turned on just by DEBUG being defined; it has to be turned on explicitly by modifying this file (which was apparently the intention anyway since it already had the #if 0 clause there, it just didn't protect itself against someone else defining DEBUG for their own purposes).


hughsando:
Thanks.

PR #207, opened 2015-03-26
Topic socket improvements

bjitivo:
Adds some improvements to the Socket class that we need for some other haxe standard library changes.

Specifically, throws EOF when EOF has occurred instead of returning null (this is more correct behavior according to the API), also throws a Blocking exception in connect for nonblocking sockets instead of an error that the haxe standard library layer misinterprets as a connection failure.  Finally, adds a "fast" version of select that doesn't churn tons of objects, which will be used by a subsequent haxe std library change.


bjitivo:
This is building now.  Any comments?  Would you be able to pull this?  We have changes that we want to submit to the haxe foundation for inclusion in the haxe standard library that require this.

The churn associated with socket select is significant and this change helps eliminate that.

bjitivo:
Re-worked the topic branch to have a much saner list of changes.

PR #208, opened 2015-03-26
Topic unload prims

bjitivo:
We found that the primitives loaded by __loadprim must be unloaded sometimes for proper shutdown.  So this code enables this.  A separate change in the haxe std library will use these functions to ensure that shutdown goes smoothly.


hughsando:
I have integrated these changes, however you must explicitly call cpp.Lib.unloadAllLibraries() yourself before Sys.exit().

PR #209, opened 2015-03-26
Topic string churn

bjitivo:
Reduce string churn by adding a comparison function to the String class that compares directly against const char * strings.  Then, don't use HX_CSTRING(xxx) all over the code to compare against static string values, since that macro creates a String class on the stack.  Avoiding this object creation reduces churn.

### DIFF
--- a/include/hxString.h
+++ b/include/hxString.h
@@ -149,6 +149,8 @@ public:
 
    inline bool operator==(const ::String &inRHS) const
                      { return length==inRHS.length && compare(inRHS)==0; }
+   inline bool operator==(const char *other) const
+      { return other ? __s ? (strcmp(other, __s) == 0) : false : !__s; }
    inline bool operator!=(const ::String &inRHS) const
                      { return length != inRHS.length || compare(inRHS)!=0; }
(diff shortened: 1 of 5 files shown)


hughsando:
HX_CSTRING should be using the inline-constructor ```inline String(const HX_CHAR *inPtr,int inLen)``` to create a stack struct that joins the constant (not copied) pointer with the length.  So there should not be any churn.  Are you detecting allocations in this case?

hughsando:
Actually, I use for example : HX_FIELD_EQ(inName,"main") in the generated code.

hughsando:
... Although I have made sure the length is correct before doing this.

bjitivo:
I can't remember the details of why I thought there was churn involved in these comparisons.  I might have been wrong.

bjitivo:
OK I think my confusion was believing that the HX_CSTRING macro was allocating the strings from the garbage collector instead of the stack.  But on re-read I believe the string objects are being allocated on stack and not using the garbage collector at all.  And furthermore, the constructor is inlined and doesn't really do anything and I expect that the compiled code would be minimal.

If you agree with the above, feel free to reject this pull request.  Thanks!

hughsando:
Cool - I think we are on the same page.

PR #210, opened 2015-03-26
Topic unordered map

bjitivo:
Use std::tr1::unordered map in most places where std::map had been used.  We found that some functionality in our application was measurably faster this way.  However, I'm not sure what the C++ compiler requirements are for using tr1 functions and whether or not you want to introduce tr1 dependencies in the code.  Perhaps I should have guarded this with a -D macro ... let me know if you'd like me to do that.


hughsando:
Constant Strings generated by haxe now have built-in hash values - this could potentially be used to speed-up the lookups - this would involve keeping the key as String, rather than std::string, and providing the hashing function.
There is also now the src/hx/Hash.h, which could potentially be coaxed into using malloc rather than GC new to create its array, and this would provide a cross-platform fast string map.  Not sure how this would compare to unordered_map.

bjitivo:
I believe the AppVeyor build is failing due to a time-out in the CI system.  Not sure how to kick it to try to get it to do a new build.

Simn:
I just told it to rebuild, let's see what happens.

bjitivo:
OK the Windows build shows that there's no tr1/unordered_map for whatever C++ compiler configuration is used on Windows.  So that pretty much kills this idea.  Feel free to close this pull request and we'll maintain this patch locally since we need it but don't care about Windows.  Thanks!

hughsando:
It may be possible to use a typedef to go faster if unordered_map is
available (gcc/clang), but fallback otherwise. My main concern is switching
between map<std::sting> and map<String>, since constructing an std::string
from a const char * will use an alloc.  If we can be sure one way or
another that only "constant" strings (HX_CSTRING, haxe compile-time
constants or strings explicitly allocated as such) are used, then we might
just be able to do a unordered_map<String>.
In cases where the map is of std::string, and order is not important (which
it is not), there should be only benefits of switching string maps so this
would be a 1-for-1 swap - if we use a typedef then there should hardly be
any changes.

If the optimization benefits need to trade a c++ alloc vs a log(N)
algorithm, it is not clear under what circumstances this is a win.  If you
can give some idea of where you can measure the performance difference, I
can have a look to see the best way of getting the speedup on as many
platforms as possible.  I'm thinking the hx/cpp/Hash.h code may be useful
because it can already use pre-generated hashes.
But I'm not that keen to try any optimization work without a clear way to
profile any improvements.

hughsando:
I'm closing this now, but some of the ideas live on in hx/Unordered.h, which is used internally for some structures where appropriate.

PR #212, opened 2015-03-26
Topic val is buffer

bjitivo:
Added val_is_buffer, which we need in our application.


hughsando:
val_is_buffer is now part for the CFFI API, since it needs a different implementation for neko.  So this should be working for you now.

PR #214, opened 2015-04-01
Fix debugger bug: when the debugger thread itself experiences a critical...

bjitivo:


bjitivo:
I don't get it.  There are no merge conflicts.

Simn:
Did you pull hxcpp locally? Maybe there was a recent change that is causing a conflict.

PR #215, opened 2015-04-01
Fix debugger bug: when the debugger thread itself experiences a critical...

bjitivo:
... error, do not attempt to break into the debugger as this will deadlock.


bjitivo:
Yeah I just don't get it.  This pull request does not contain any merge conflicts, period.

bjitivo:
Ah I see it now.  I don't understand how this happens, but I see it.

bjitivo:
OK fixed it.

bjitivo:
So the CI is not starting though?

bjitivo:
OK it's starting now.

PR #220, opened 2015-04-16
use OBJDIR from env to configure .o output dir

yanhick:
This is a change we made to hxcpp to allow specifying the .o output dir. We can change the `BUILDDIR` name to something else.

It's only implemented for the targets we use (linux and macos), but I can implement it for the other targets if you're interested in this change.


hughsando:
Do you still want this?
The HXCPP_COMPILE_CACHE means that most of the obj files are elsewhere now.

hughsando:
A bit old now.  I can look at it again if needed.

PR #221, opened 2015-04-16
make DESTDIR configurable

yanhick:
For our builds, we need to be able to choose the build output system wide, so we made it possible to use absolute paths.


hughsando:
hxcpp does not use these binaries anymore.

PR #222, opened 2015-04-16
added dockerfile and run test script

yanhick:
This is the Dockerfile I use to run the tests for Linux locally.

With Docker installed, you can use it like this:

```
docker build -t hxcpp_tests .
```

This creates an image named `hxcpp_tests` from the Dockerfile.

```
docker run -v ${PWD}:/hxcpp -t hxcpp_tests
```

This starts a container which test the code mounted with the `-v` option

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,46 @@
+# Use same distribution as Travis.CI
+FROM ubuntu:12.04
+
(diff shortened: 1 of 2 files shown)


hughsando:
Is this something you are still interested in?

hughsando:
closing because of age

PR #224, opened 2015-04-17
Added missing #includes to make it compatible with libc++

martinfinke:
String.cpp uses `atoi` (from `<cstdlib>`)
Lib.cpp uses `getenv` (from `<cstdlib>`)


hughsando:
These files now include "stdlib.h".  I'm going to close this one, but if you are getting errors/warnings you can re-open it.

martinfinke:
Oh ok, I haven't used this in more than a year.

PR #226, opened 2015-04-29
Resolve haxelib directory in the following order

dpeek:
1. `./.haxelib` - local haxelib repository
2. `$HAXELIB_PATH` - environment variable supported by haxelib*
3. `~/.haxelib` - globally defined haxelib repository
4. `$HAXEPATH/lib`
5. `/etc/.haxepath` - apparently this is a thing too
6. `$DEFAULT_HAXEPATH/lib` - assume default install location

The issue I have currently is that `$HAXEPATH` is set but my haxelib repository is in another castle.

* see https://github.com/HaxeFoundation/haxelib/blob/4fefbd17a2d9a98200b621de801018af3896d68a/src/tools/haxelib/Main.hx#L949

### DIFF
--- a/src/hx/Lib.cpp
+++ b/src/hx/Lib.cpp
@@ -244,21 +244,29 @@ String FindHaxelib(String inLib)
 
    if (haxepath.length==0)
    {
+      haxepath = GetEnv("HAXELIB_PATH");
+      if (loadDebug)
+         printf("HAXELIB_PATH env:%s\n", haxepath.__s);
+   }
+
+   if (haxepath.length==0)
+   {
+       #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


dpeek:
Hoping this can make it into 3.2 – the current logic ignores what was set by `haxelib setup` if the user has set `HAXEPATH` (which people that switch Haxe versions a lot might do, ie. with https://github.com/dpeek/hvm).

hughsando:
Well, you have sort of missed what I was thinking would be the release to coincide with the 3.2 release.  But never fear, there will be a new hxcpp release pretty soon I should think.

dpeek:
Of course, I forgot the runtime is versioned independently :)

See you at WWX!

PR #235, opened 2015-05-29
Update msvc-setup.bat to include using VS2015 Toolchain

benwart:


hughsando:
Have not installed vs2015 but this looks nice.

PR #236, opened 2015-05-30
Add socket_send_to and socket_recv_from for UdpSocket

Meoo:


Meoo:
Related haxe PR : https://github.com/HaxeFoundation/haxe/pull/4267

ncannasse:
@hughsando we have these in neko as well

@Meoo are they backported from here ? also, seems like you will lock the GC if you have block_error

Meoo:
I copied the the send / receive functions of the same file and used neko's socket.cpp as a reference.
block_error calls gc_exit_blocking, so it shouldn't lock the GC.

hughsando:
send_to and recv_from have been merged

PR #237, opened 2015-06-01
Telemetry Support

jgranick:
This is a list of commits from @jcward, author of hxscout. Looking at the "Files changed" tab, most of these changes look sound and straightforward, though the array changes are more difficult to follow.

Since this is guarded behind an "HXCPP_TELEMETRY" define, I think it may be pretty safe to do, but @hughsando I thought you might like to take a look so that we can have profiler support built-in :smile:

Thanks, we can also use this pull as a place to discuss


jcward:
Hi Joshua,

I'm not sure why the Array file looks totally redone - probably bad whitespace diff'ing. I did have to merge that file in this commit: ec5a1b6

I haven't changed my fork of hxcpp in a while. I do certainly think there were be many fixes to come, but if you're happy with seemingly-stable-but-beta quality behind an opt-in compiler switch, then it's fine with me.

I should probably make a clean-up pass, there are a few blobs of commented out debug / old code.

Other than that, there's a few things I have questions about which @hughsando perhaps could weigh in on:

- I think the only thing that's not protected with HXCPP_TELEMETRY is a slight refactor of [this function](https://github.com/HaxeFoundation/hxcpp/pull/237/files#diff-4483855aa53ab4456a6e9121380a6448R3321), which I called from [here](https://github.com/HaxeFoundation/hxcpp/pull/237/files#diff-9169d4f7b906cd2fb90ee9ac7d198d5cR2724) at one point but seem to have since removed that.
- I'm not sure that using `extern` is the right thing to do e.g. [here](https://github.com/HaxeFoundation/hxcpp/pull/237/files#diff-33ad85f9be102789a9924b63a3c9d0ccR9) and [here](https://github.com/HaxeFoundation/hxcpp/pull/237/files#diff-422f532e532b78bf183a82ad6750f268R5) and more -- I'm not much of a cpp programmer. The definitions might should be included via a `.h` file.
- Telemetry uses object ID's (a 32-bit int representation of their 32/64-bit addresses, I think). I'm not really sure the [hash computation function](https://github.com/HaxeFoundation/hxcpp/pull/237/files#diff-9169d4f7b906cd2fb90ee9ac7d198d5cR336) is 100% bullet proof / collision safe, but I based it on an [existing function in GCInternal](https://github.com/jcward/hxcpp/blob/master/src/hx/GCInternal.cpp#L3471-L3481).

I'll add any other notes as I think of them.

Cheers,
-Jeff

jcward:
FYI, I'm working on that clean-up commit today.

jcward:
Pushed [clean-up commit](https://github.com/jcward/hxcpp/commit/0fd8a17f0c150e742ba4666c4614de5c67b12786) to my integration branch.

hughsando:
Hi.  This PR kind of got mixed up with wwx, and I've been busy since.
I have a few minor issues/ideas about some things. Can you just use a "void *" instead of an int for objectId ?  If these are just keys in a hash, then it should be ok.  There could be an issue here either way when the gc works in "moving mode".  This is a bit unusual, so it might be enough just to generate a compile-time error if both telemetry and moving are defined.
I was also thinking of changing  ExternalPrimitive to have the actual name of the primitive, which might cause an issue with the "extern" logic - but we can cross that bridge when we come to it.
I was using the "__hxcpp" style names for code that can be called from haxe code - for internal code, you can put it in the hx:: namespace if you like.
I'm thinking it might also be nice to split up Debug by moving the CallStack to a separate header, but this can wait.
So I'm inclined just to pull this request and deal with these other points when motivation strikes.  However, it is a lot of commits.  Do you think you can squash these into a single commit rebased of the current hxcpp head?

jcward:
This can be cancelled and closed, superseded by #243

PR #242, opened 2015-06-11
Fixed hxcpp debugger breakpoints by disabling HXCPP_DEBUG_HASHES.

bjitivo:


bjitivo:
I was pretty sure that HXCPP_DEBUG_HASHES broke debugger breakpoints but since nobody had reported it I didn't submit a pull request about it.  We are using a local version of hxcpp that has HXCPP_DEBUG_HASHES commented out like in this pull request, which is why debugger breakpoints are working for us.  I have tried a test program with Haxe 3.2 and stock hxcpp (3.2.102) and found that debugger breakpoints do not work, so in fact this is a real problem, hence this pull request.

bjitivo:
AppVeyor build failure is due to AppVeyor flakiness, which seems to be a persistent problem.

andyli:
I've just restarted the build. Hopefully it will be fine this time.

bjitivo:
Thank you for kicking AppVeyor Andy.

I'd love to hear if Hugh or anyone could verify that breakpoints don't work with HXCPP_DEBUG_HASHES enabled and that this change fixes that.  Note that breakpoints at the first line of a function work regardless so to test this you'll have to set a breakpoint deep in a function.

hughsando:
I can verify the debug hashes seem to be a problem - i'll look into it a bit more.

hughsando:
I think I have fixed it - like file+line breakpoints were hashing in the line number - it should just be the file name.  A small change if you want to test it in your code base.

bjitivo:
Sure, I'm happy to try your fix, not sure how to get it though ... my github-fu is weak ...

hughsando:
I would say just manually merge it - it is only one word:
https://github.com/HaxeFoundation/hxcpp/commit/c692bba4fbb6404273b65c2239bb299b0bd914e6

bjitivo:
I can confirm that your change fixes the issue.  So I withdraw this pull request and ask that you kindly make your change into the master hxcpp (and if possible, get the fix into the hxcpp release haxelib, so that people using hxcpp will have the fix).  Thank you!

PR #245, opened 2015-06-19
[objc] Create Dynamic wrapper around Objective-C types

waneck:
Related to HaxeFoundation/haxe#4350 : this adds an implicit operator
to convert from Dynamic to NSObject * and the other way around. The
changes are all guarded through the `-D objc` Haxe define

### DIFF
--- a/include/Dynamic.h
+++ b/include/Dynamic.h
@@ -30,7 +30,7 @@ public:
    explicit Dynamic(const HX_CHAR *inStr);
 #ifdef __OBJC__
 #ifdef HXCPP_OBJC
-   Dynamic(const NSObject *inObjc);
+   Dynamic(const id inObjc);
 #endif
 #endif
 
@@ -58,7 +58,7 @@ public:
    inline operator bool() const { return mPtr && mPtr->__ToInt(); }
 #ifdef __OBJC__
 #ifdef HXCPP_OBJC
-   inline operator NSObject * () const { return mPtr ? (NSObject *)mPtr->__GetHandle() : 0; }
+   inline operator id() const { return mPtr ? (id)mPtr->__GetHandle() : 0; }
 #endif
 #endif
    inline bool operator !() const { return !mPtr || !mPtr->__ToInt(); }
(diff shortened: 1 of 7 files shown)


waneck:
Some implementation details:

* This will wrap any `NSObject *`-derived type; Because of variance, an additional cast (from `NSObject *` to the actual casted type) was inserted on the `gencpp` side
* Since apparently hxcpp does not call C++ destructors for garbage-collected types ( @hughsando , is that right? ), ARC doesn't seem to be able to won't work on hxcpp; Having this said, when boxed the types will always call `retain` on the object, and a finalizer will call `release`
* I've used `vtAbstractBase + 4` for its kind type. Strangely, it seems that [this will conflict with the CFFI](https://github.com/HaxeFoundation/hxcpp/blob/master/src/hx/CFFI.cpp#L107) . However this is already the case for `cpp:Pointer` and `cpp:Struct`, so I decided not to touch on the CFFI code just yet

hughsando:
Yes - no destructors, so finalizers are your best bet.
It might be possible to use one type for stack variables, and another type
for member/arrays and use ARC on the stack and some kind of "AutoFinalizer"
on the heap (eg, always box these).
I think if you bump the 'sgKinds' initial value, you can add your own
pre-allocated 'kind' for abstracts.  I could maybe refactor this a bit.

waneck:
@hughsando , thanks! I've just added a commit which allows objective-c protocols to work too; Also I've just realized that the very same place I linked about the `sgKinds` at `CFFI.cpp` already has the other cpp kinds. I guess I was too tired to miss that part, sorry ;)
So I've now added the objective-c box type to `sgKinds` as well.

Thanks!

hughsando:
Another idea I had was to tag certain classes with something like @:delete,
and in this case automatically adds a finalizer for the class and call the
c++ "delete" on the class from it (which you could perhaps override with a
@:destructor meta function.  Classes with objc references could then clean
up as required.  This would potentially allow you to have 1 finalizer
instead of many and avoid extra objects.  The constructor basically memsets
the struct to 0, but some kind of @:constructed meta may allow you to have
things like std::strings in a class without boxing.

waneck:
That'd be a great addition. It would allow ARC to "just work" on hxcpp, it seems. Also there is still the lingering problem with regards to objective-C references inside a hxcpp class, even if ARC isn't able to work for some reason - so we can release the retained references - manually, if needed

PR #248, opened 2015-07-01
Fixes crashes when compiling for Blackberry 10.3

DanielUranga:
This PR setups hxcpp to use qcc instead of g++ when compiling for Blackberry. Modified to match the compiler calls that Blackberry Momentics IDE does as best as possible.

Some modifications to the source files of hxcpp were made too, mostly adding:
```
#if defined(BLACKBERRY)
using namespace std;
#endif
```
to files that required it.

### DIFF
--- a/include/hx/CFFI.h
+++ b/include/hx/CFFI.h
@@ -16,12 +16,9 @@ typedef struct _buffer  *buffer;
 #include <stdio.h>
 #include <stdlib.h>
 #include <memory.h>
-
-
-
-
-
-
+#if defined(BLACKBERRY)
+using namespace std;
+#endif
 // --- Register functions (primitives) ----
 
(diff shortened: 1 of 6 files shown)


hughsando:
@jgranick - do you have an opinion on this?

jgranick:
Let me give it a try, and see how it works for me :smile:

jgranick:
It looks good. `qcc` is a front-end to `g++`, not unlike the iOS compiler tools. It seems like all the BlackBerry samples use it, and it enables some specific, extra features that might not be available calling `g++` directly.

It works here using an old BlackBerry native SDK, with the addition of a `using namespace std` for TinyXML. I think this is a good pull if you are happy with it :smile:

PR #249, opened 2015-07-03
Support for VS2015 64 Toolchain

DanielUranga:
The -FS option (Force Synchronous PDB Writes) is needed in order to allow multiple compile threads.
https://msdn.microsoft.com/en-us/library/dn502518.aspx

### DIFF
--- a/toolchain/msvc-toolchain.xml
+++ b/toolchain/msvc-toolchain.xml
@@ -35,6 +35,7 @@
 <set name="MACHINE" value="x64" if="HXCPP_M64" />
 
 <compiler id="MSVC" exe="cl.exe" if="windows">
+  <flag value="-FS"/>
   <flag value="-nologo"/>
   <flag value="-Od" if="debug"/>
   <!-- -O2 is back ! -->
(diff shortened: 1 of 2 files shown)


hughsando:
I think the FS option is only needed for debug (-Zi) builds, and there is something in there for this, via the HXCPP_FORCE_PDB_SERVER setting in debug mode.  Maybe this needs to be set for vs2015?

DanielUranga:
If I remove the ```-FS``` I get and an error ```C1041``` from ```link.exe``` when linking ```lime.ndll```.
Maybe this: ```<unset name="HXCPP_FORCE_PDB_SERVER" if="HXCPP_COMPILE_CACHE" />``` should only be done on versions of MSVC older than 2015?

hughsando:
I have added these changes, although I had to use 'vcvarsx86_amd64' to get at the batch file.
I'm still not sure how the build is using a pdb while not using the -FS flag, but I have moved this flag, since I don't think it will cause any harm if no pdb is used.

PR #260, opened 2015-07-26
Fixed setsockopt SO_NOSIGPIPE for Mac

jcward:
hxtelemetry apps are crashing with SIGPIPE when hxscout is shut down. See [hxscout issue #60][https://github.com/jcward/hxScout/issues/60].

With the previous arguments, setsockopt was returning -1, error. I found this format on Stackoverflow, which returns 0 (success) and this causes my hxtelemetry apps not to crash when hxscout is shut down (closing the socket from the far end) on Mac.

I also wonder why the MSG_NOSIGNAL isn't stopping SIGPIPE on Mac/Win -- should it be? Or is #define MSG_NOSIGNAL = 0 disabling this behavior for win and mac? (line 32) I still need a fix for Windows...

### DIFF
--- a/project/libs/std/Socket.cpp
+++ b/project/libs/std/Socket.cpp
@@ -131,7 +131,8 @@ static value socket_new( value udp ) {
 	if( s == INVALID_SOCKET )
 		return alloc_null();
 #	ifdef NEKO_MAC
-	setsockopt(s,SOL_SOCKET,SO_NOSIGPIPE,NULL,0);
+	int set = 1;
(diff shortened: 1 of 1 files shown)


jcward:
@hughsando - any comment on this? Need any further info?

The [OSX docs for setsockopt](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man2/getsockopt.2.html) state:

    If no option value is to be supplied or returned, option_value may be NULL.

But it's a confusing statement since the get/set docs are combined.

hughsando:
Seems reasonable!

PR #272, opened 2015-08-25
Fixed telemetry memory leak (stats were not being deleted properly)

jcward:
Fix for telemetry stats leaking (deleting std::vectors inside each TelemetryFrame)

### DIFF
--- a/src/hx/Debug.cpp
+++ b/src/hx/Debug.cpp
@@ -471,8 +471,17 @@ public:
         gStashMutex.Unlock();
         return 0;
       }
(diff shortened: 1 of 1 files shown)


jcward:
@hughsando - once this is fixed and included with version 3.x.xxx, I'd also like to add a compile-time warning to hxtelemetry `#if hxcpp < 3.x.xxx` -- "Warning - telemetry memory leak in hxcpp < 3.x.xxx" -- do you think that's reasonable?

PR #275, opened 2015-09-01
add CFFIPrime test for zero argument function

nadako:
see https://github.com/HaxeFoundation/haxe/pull/4516


hughsando:
I have "getNullString" now that should test this.

Issue #280, opened 2015-09-09
MSVC cl.exe version detection fails on non-English environments


mrcdk:
This block fails to detect the correct `cl.exe` version:
https://github.com/HaxeFoundation/hxcpp/blob/master/tools/hxcpp/Setup.hx#L536-L561

Because on non-English environments the output of `cl.exe` is localized so, in Spanish, the output is:
`Compilador de optimización de C/C++ de Microsoft (R) versión 19.00.23026 para x86`

Based on this: http://stackoverflow.com/a/1233332/3750270 maybe changing the regex to `~/(\d{2})\.\d+/` would do the job.

PR #283, opened 2015-09-16
Improve msvc cl.exe version checking for non-English environments

mrcdk:
Related issue https://github.com/HaxeFoundation/hxcpp/issues/280


mrcdk:
The errors don't seem to be related to this PR.

hughsando:
I'm not 100% sure if this is the right regex - but it looks pretty good.  I guess we'll find out soon enough.
Thanks

PR #284, opened 2015-09-17
Added support to search cross-compiler to build Windows-32bit from Linux-64bit

jobs-git:
Added support to search on cross-compiler to build Windows-32bit executable from Linux-64bit, recompiling of hxcpp is required to take effect

### DIFF
--- a/tools/hxcpp/Setup.hx
+++ b/tools/hxcpp/Setup.hx
@@ -114,7 +114,7 @@ class Setup
             return;
          }
 
-         var guesses = ["c:/MinGW"];
+         var guesses = ["c:/MinGW","/usr/i686-w64-mingw32"];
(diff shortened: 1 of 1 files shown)


hughsando:
I think you can do this without changing hxcpp via the "MINGW_ROOT" variable.
You can set this like:   haxe .... -D MINGW_ROOT="/usr/i686-w64-mingw32", or set it in your ~/.hxcpp_config.xml file which is designed for machine-wide settings like this one.

jobs-git:
Ok, I will just add -D Ming_root on WindowsPlatform.hx of Lime.

jobs-git:
Ok, I will just add -D Ming_root on WindowsPlatform.hx of Lime.

PR #297, opened 2015-10-02
tvOS support

tanis2000:
This PR adds support for a new tvOS target in HXCPP


jgranick:
This looks pretty good to me. What do you think @hughsando? :smile:

Simn:
In general we should have some maintenance policy for new hxcpp targets. In particular there should be someone we can @-mention in case of issue reports that are specific to a target, kind of like e.g. travis does it.

jgranick:
I usually add new targets to HXCPP, but I thought I would ask Hugh in case he had some red flag or existing tvOS ideas in mind

Simn:
Sure but you already maintain a million other things, so it would be cool if @tanis2000 agreed to helping out with tvOS-specific issues if any come up.

jgranick:
Ah, gotcha

tanis2000:
You can count on me. I plan on releasing some games on that platform ASAP so I'd be willing to maintain it. Of course my time is limited and I have a million other things to do, but if there's not too much pressure then it's fine by me.

hughsando:
Have not looked too closely, but it seems very good.  But I probably have a
bit of work to do to make sure the build server generates these libraries
automatically.

On Sat, Oct 3, 2015 at 3:18 AM, Valerio Santinelli <notifications@github.com

hughsando:
Thanks - you saved me quite a bit of work.

tanis2000:
You're welcome :)

PR #298, opened 2015-10-03
Fix problematic cross compilation of Windows exe in Linux host

jobs-git:
When compiling (From Linux host) using

    haxelib run hxcpp Build.xml -Dwindows -Dmingw -DMINGW_ROOT=/usr/i686-w64-mingw32

hxcpp tries to compile Linux executable, which should not be the case since we implicitly indicated windows

This happens because the original code below is not enabled for cross compilation

    else if ( (new EReg("linux","i")).match(os) )
    {
        set64(defines,m64);
        defines.set("toolchain","linux");
        defines.set("linux","linux");
        defines.set("BINDIR", m64 ? "Linux64":"Linux");
    }

The above pull fixes this problem.


hughsando:
I merged the xcompile toolchain patch.  Do you still want this one too?

jobs-git:
yes, this should be merge so that the toolchain can work, i will resolve the conflict for this one.

jobs-git:
I will be opening a cleaner pull request, I will now close this.

PR #304, opened 2015-10-06
Defaults Mingw toolchain to g++.exe when not cross-compiling

jobs-git:
Windows users can now compile Win32 app using

    hxcpp Build.xml

While Linux users can do the same using

    hxcpp Build.xml -Dwindows

NOTE:

Windows users who intends to use Mingw as their compiler could install it from this site:

    http://sourceforge.net/projects/mingw-w64/

Linux users on the other hand could do:

    sudo apt-get install mingw-w64

or

    sudo yum install mingw32 mingw32-gcc-c++ mingw32-headers mingw32-crt mingw-winpthreads mingw32-cpp mingw32-filesystem

For a detailed discussion, please refer to 

HaxeFoundation/hxcpp#294

### DIFF
--- a/toolchain/mingw-toolchain.xml
+++ b/toolchain/mingw-toolchain.xml
@@ -11,6 +11,8 @@
    </section>
 </section>
 
+<set name="CXX" value="g++.exe" unless="xcompile"/>
+
 <setup name="mingw" />
 
 <path name="${MINGW_ROOT}/bin"/>


jgranick:
After the mingw32 changes, I believe that users on Windows (with no compiler) receive errors about mingw32. Is there some way to show an error about not having Visual Studio C++ installed (or a C++ compiler installed) in this case? I'm getting reports lately about users expecting mingw32 is required for Haxe Windows builds, I think it should be optional, and not default.

Does that sound correct? Thanks

jobs-git:
I am thinking It may be from setup.hx in hxcpp folder, I get the same error even before this patches.

There are several options we can resolve this:

1.) Defaults compiler to VS C++ compiler, return string VS C++ error when not found.
2.) Defaults compiler to Mingw, return string mingw error when not found. (Current issue?)
3.) Looks for either of the compilers, return string when all compilers where missing.

I am thinking of doing 3.) as pull request, but I have explored very little of hxcpp source files.

PR #306, opened 2015-10-08
Support for tvOS beta3

tanis2000:
This only adds the needed conditionals to turn on bitcode creation. It's mandatory starting from beta3


nensanders:
It would be very convenient for us to have this merge released soon.
For us its working on the simulator as well as on our AppleTV dev kit machine. Thanks for the contribution tanis2000. (https://github.com/gameduell/duellbuildtvos)

PR #309, opened 2015-10-08
 Fix: Initialized xcompile toolchain when cross-compiling

jobs-git:
ABOUT THE PULL:

When in Linux, the old code will default you to compiling Linux binaries, what ever -Dwindows we do.

This is because the old code is not configured to enable cross-compilation yet.

This pull was offered to ensure that cross compiling from Linux will call on the xcompile toolchain.

Here, only the Linux part of the Condition statement was updated, Windows and Mac users will get the expected action of hxcpp.

To discuss further, the code will ensure to run the xcompile when we attempt to build a windows application in a Linux host by submitting the

    -Dwindows

to the compilation. Else, if we did not include it, the code will resume to compile a Linux binary, thus, we recover the same behavior as the old code.

FULLY WORKING ON:

1.) Fedora
2.) Ubuntu

This goes in complement to the following

HaxeFoundation/hxcpp#294
HaxeFoundation/hxcpp#304

NOTE: xcompile, within the mingw-toolchain.xml. is activated only when attempting to cross-compile (One Platform to  Another Platform).

### DIFF
--- a/toolchain/mingw-toolchain.xml
+++ b/toolchain/mingw-toolchain.xml
@@ -5,6 +5,7 @@
 <section if="xcompile" >
    <section if="linux_host">
       <set name="CXX" value="i686-w64-mingw32-g++"/>
+      <set name="MINGW_ROOT" value="/usr/i686-w64-mingw32"/>
    </section>
    <section if="windows_host">
       <set name="CXX" value="g++.exe" />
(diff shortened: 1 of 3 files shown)


jcward:
Wow, looks interesting. Are there package requirements (mingw?) Does it nicely report missing packages? I'll test this soon.

jobs-git:
@jcward yes there is, those were detailed here https://github.com/HaxeFoundation/hxcpp/pull/294

jobs-git:
@thomasuster yeah, I believe it is much more cleaner that way. but I think it may deviate a little bit from the original coding style of hxcpp.

PR #311, opened 2015-10-11
Only use fast path if slot is less than 64

waneck:
The trick mentioned there only works for < 64 slots. I'm working on integrating with a big application that actually uses those 64 slots before hxcpp gets initialized, and got some very hard to track errors.

It might be possible to make the code work on slots bigger than 64 using http://cbloomrants.blogspot.com/2013/09/09-18-13-fast-tls-on-windows.html. I really didn't want to make a PR of code that I don't understand

I'm curious however if this mode is faster than `__declspec(thread)`. I thought this was the faster way to support TLS.

### DIFF
--- a/include/hx/Tls.h
+++ b/include/hx/Tls.h
@@ -57,7 +57,10 @@
          DATA **extra = (DATA **)(__readfsdword(kTibExtraTlsOffset));
          return extra[mFastOffset];
          #elif (_MSC_VER >= 1400) & !defined(HXCPP_DEBUG)// 64 bit version...
-         return (DATA *)__readgsqword(mFastOffset);
+         if (mSlot < 64)
+           return (DATA *)__readgsqword(mFastOffset);
+         else
+           return (DATA *)TlsGetValue(mSlot);
          #else
          return (DATA *)TlsGetValue(mSlot);
          #endif


ncannasse:
There seems to have some restrictions on __declspec(thread) with dynamic loading, which is required for NDLLs

waneck:
But then shouldn't this code only be active if DLL_EXPORT is defined? Tls is used on the GC a lot, and hxcpp only supports dynamic loading of its own executables if some defines are set (I think DLL_EXPORT is one of them)

hughsando:
Are you using 64 bit?  I thought the 64 slots was not a problem on 64 bits
- although the reference you point to suggests it is.
Hxcpp can generate libs which can get linked into exe or dlls, so things
aren't so simple.

waneck:
Yeah, I'm compiling to 64-bit.

waneck:
I understand about the libs. It might be nice to have a flag that makes TLS use gcc's `__thread` and msvc's `declspec(thread)` though, As far as I know, they use some link-time optimizations which really aren't available otherwise.
Also the link I posted seems to indicate this is a problem only on pre-vista windows. I wouldn't mind if XP were available only if a specific flag is defined

waneck:
@hughsando any news about this?

waneck:
It would be great to have this fixed for the next hxcpp version. This problem comes up frequently when running Unreal.hx on Windows (Unreal is 64-bit)

hughsando:
I think the TlsGetValue is slower than __declspec, which is slower then the direct register access.
This fix keeps the speed good for the "average" use, and works for the more unusual case with external TLS slots, which is nice.
I would like to find the equivalent fast version for 64bit/64 slot, but maybe that can wait.

PR #320, opened 2015-10-22
Enable bitcode for ios builds

jeremyfa:
I applied changes similar to what has been done for tvos.

Even if we can still submit apps on iOS without bitcode enabled, Xcode enables bitcode by default on any new ios project.

Instead of having to disable it, it is probably wiser to just make it the default on hxcpp as well.

Let me know if the changes are correct or not.


thomasuster:
@jeremyfa Would you test your changes on ios9 device?

jeremyfa:
@thomasuster I did run on an iPhone 5s with iOS 9.1 an xcode project that includes recompiled hxcpp std/regexp/... static libraries. I could enable bitcode on the xcode project and generate an archive that can have bitcode as well. So far it was allright.

That said there is one more test that I would like to do: try the changes with something like openfl or snow, with bitcode enabled on the xcode project. I will probably try with snow tomorrow.

thomasuster:
@jeremyfa Okay awesome. I'm sure Hugh would appreciate that test. I can also test this with Pakka Pets  (NME) once you're confident in the change.

jeremyfa:
@thomasuster Recompiled snow binaries from modified HXCPP, then just exported an ipa with bitcode enabled from a luxe project. Worked fine! I believe the change is ok, and should not create any problem with ios projects that disable bitcode as well.

thomasuster:
Awesome!

hughsando:
I wonder if the flag would be better in the toolchain files (iphoneos-toolchain.xml/iphonesim-toolchain.xml) .  In builder.hx, the flag will go on for the stdlibs (regexp, zlib etc) but will not go on for the main haxe code.
Also, I guess it needs xcode 7.1 release, so I might need to detect this.

jeremyfa:
The flags are currently both in the toolchain files and builder.hx. I applied similar changes to what has been done for tvos: https://github.com/HaxeFoundation/hxcpp/commit/ab58d23ce8337290a78efae088272212c247263e

I guess if the flags are not required in builder.hx, we should also remove them for tvos.

And yes, it might be cleaner to detect xcode version (version >= 7.0?).

hughsando:
I have added the flags to the toolchain file.  Don't think I need them on the simulator.
I still might need a way to disable bitcode for older systems.

PR #323, opened 2015-10-26
Blackberry compile fix

DanielUranga:
Move "#include <string.h>" from GcRegCapture.cpp to GcRegCapture.h. This is needed in order to be able to compile on Blackberry. Other targets will keep working ok.


hughsando:
This seems a little odd - what error were you getting?

hughsando:
Ok, size_t - I see the other issue now.

hughsando:
I have just pushed a change to GcRegCapture.cpp which should be better on windows too.  See if this fixes the issue.

DanielUranga:
It fixes the issue :+1:

PR #330, opened 2015-11-10
add Array.reduce (see HaxeFoundation/haxe#4637)

nadako:


nadako:
https://github.com/HaxeFoundation/haxe/issues/4637#issuecomment-194602553

PR #334, opened 2015-11-18
added HXCPP_OPTIMIZE_FOR_SIZE

sebthom:
When creating non-gaming apps for mobile devices optimization for APK size may be desirable. Specifying the compiler flags **-Os**, **-fno-asynchronous-unwind-tables** and the linker flag **-Wl,--gc-sections** reduced the size of a test haxeui APK by 25%.

The compiler flag -Os is described in the gcc manual with: *-Os enables all -O2 optimizations that do not typically increase code size.*

With this pull request, the new **HXCPP_OPTIMIZE_FOR_SIZE** switch will be introduced for the Android toolchain, which - if specified - enables the mentioned compiler/linker flags.

openfl projects can now for example use:

    <haxedef name="HXCPP_OPTIMIZE_FOR_SIZE" if="android" />

### DIFF
--- a/toolchain/android-toolchain.xml
+++ b/toolchain/android-toolchain.xml
@@ -123,7 +123,12 @@
  <flag value="${ANDROID_PLATFORM_DEFINE}"/>
  <!-- todo <flag value="-Werror"/> -->
  <flag value="-Wa,--noexecstack"/>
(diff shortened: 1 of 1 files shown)


hughsando:
Would be nice to have this for the other targets too

PR #336, opened 2015-11-21
Update WinRT to Windows 10 UWP

madrazo:
Hi. 

Here is our ***awesome*** contribution to make HXCPP work with Windows Universal Platform (aka WinRT for Windows 10), which includes XBOX ONE. 

-I am considering Windows 8 WinRT deprecated, but we can add flags if needed. 
-SQLLite still not compiling.
-There seems to be an issue with threading, thus "HAS_ATOMIC" is disabled.
-Had to add a separate msvc-winrt-setup.bat file to set environment variables.
-Fix DLL loading

Please give it a try ;)

### DIFF
--- a/include/hx/Thread.h
+++ b/include/hx/Thread.h
@@ -48,7 +48,7 @@ inline int HxAtomicInc(volatile int *ioWhere)
 inline int HxAtomicDec(volatile int *ioWhere)
    { return __atomic_dec(ioWhere); }
 
-#elif defined(HX_WINDOWS)
+#elif defined(HX_WINDOWS) && !defined(HX_WINRT) //winrt test
 
 inline bool HxAtomicExchangeIf(int inTest, int inNewVal,volatile int *ioWhere)
    { return InterlockedCompareExchange((volatile LONG *)ioWhere, inNewVal, inTest)==inTest; }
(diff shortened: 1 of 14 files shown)


madrazo:
i have added some compile fixes

madrazo:
I am using lime (legacy), to compile hxcpp:
    lime rebuild hxcpp windows -Dwinrt -clean
And the run DisplayingABitmap sample
    lime test project.xml windows -Dwinrt -clean

hughsando:
Looks good.  I think we should probably use const_cast always - no need to depend on WINRT.
Can this be tested from a command-line, or do you need the full app thing?  It would be nice to add a CI test so I do not accidentally break WINNT when building something.
I'm going to close an reopen so the CI test are run again since the telemetry fix changed some of the common files.

madrazo:
Cool. I have only tried the DisplayingABitmap sample but I will try to do a text only sample. Also it needs some packaging after compiling.
maybe const_cast<HX_CHAR *> instead of const_cast<char *>
Also note that I could not load .ndll files, only .dll files (even if contents are the same), but not sure if its just me.

madrazo:
I have seen that it is possible to create a regular windows console program that calls winrt code, which may work but have not test it (project is not updated for vs2015 uwp) https://visualstudiogallery.msdn.microsoft.com/3c68246b-8103-467c-9557-894b1f837591

Also found here 3 ways to create the appx's: manifoldjs, visual studio, and makeappx command https://github.com/manifoldjs/ManifoldJS/blob/6eb87fc4759157459af974951127730024abe044/docs/Windows10-next-steps.md so it can be done without visual studio, similar to Android

hughsando:
I'm wondering if this is better left up to the framework - eg lime/nme/snow
where they take the exe produced by hxcpp and add the manifest to it.
This is pretty much how the mobile targets work.  This allows you to reuse
configuration bits, like icon, app id, assets packing etc.

PR #364, opened 2016-01-16
Automatically default to the most compatible compiler in Windows (MSVC)

jobs-git:
If users are knowledgeable on using Mingw and are aware of some compatibility issue that may arise, they typically attempt to do Mingw compilation through `-Dmingw`, `HXCPP_MINGW` or `minimingw`.

Else, for all other users will default to the most compatible compiler for windows through MSVC.


hughsando:
Yes, I think your reasoning is pretty sound.  We will see if anyone complains.

PR #370, opened 2016-02-10
tools: add <linker> needsQuotes attribute

james4k:
Certain toolchains do not support quotes in the all_objs file.

### DIFF
--- a/tools/hxcpp/BuildTool.hx
+++ b/tools/hxcpp/BuildTool.hx
@@ -636,7 +636,11 @@ class BuildTool
                case "ranlib" : l.mRanLib = (substitute(el.att.name));
                case "recreate" : l.mRecreate = (substitute(el.att.value)) != "";
                case "expandAr" : l.mExpandArchives = substitute(el.att.value) != "";
-               case "fromfile" : l.mFromFile = (substitute(el.att.value));
+               case "fromfile" :
+                  if (el.has.value)
+                     l.mFromFile = substitute(el.att.value);
+                  if (el.has.needsQuotes)
+                     l.mFromFileNeedsQuotes = parseBool(substitute(el.att.needsQuotes));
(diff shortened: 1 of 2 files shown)


madrazo:
Hi @james4k . I want to try this. Is it used like: `<fromfile value="@" needsQuotes="0"/>` ?
Long ago I modified this file to avoid quotes entirely and I had no problem, but I suppose that some toolchains need them.

james4k:
Yes, or needsQuotes="false".

I might rename it to "noQuotes" and invert the logic, so that the default
value is false.
On Apr 7, 2016 9:50 AM, "Carlos Madrazo" <notifications@github.com> wrote:

Hi @james4k <https://github.com/james4k> . I want to try this. Is it used
like: <fromfile value="@" needsQuotes="0"/> ?
Long ago I modified this file to avoid quotes entirely and I had no
problem, but I suppose that some toolchains need them.

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/HaxeFoundation/hxcpp/pull/370#issuecomment-206939153>

james4k:
Or maybe the logic should stay the same and just name the attribute "quotes".

madrazo:
@james4k Thanks. It works as intended. For me "needsQuotes" is OK, since the reason of failing is that the toolchain puts the extra quotes already (so the quotes from hxcpp are no needed). 
@hughsando please consider merging this as it has no effect on current toolchains. Thanks.

hughsando:
Nice, thanks.

madrazo:
Hi. I think that this could be used for "flags" as well.  If I try to add a 
< flag value='-MYFLAG"C:\PATH WITH SPACES"' /> 
it writes double-double quotes, like "-MYFLAG"C:\PATH WITH SPACES""

PR #372, opened 2016-02-12
include string.h instead of memory.h

james4k:
It appears that memory.h is included for the memcpy function, but the C standard
specifies that memcpy is defined in the string.h header.

This fixes a build issue on proprietary toolchain(s).


james4k:
The failed check btw appears to be fixed by https://github.com/HaxeFoundation/hxcpp/commit/53bdca01aebb257d9f2766980dec937c35171dde

andyli:
Would you rebase your commit and then force push to your PR branch?
It will let the CIs rebuild.

PR #373, opened 2016-02-12
Ignore library loading code when STATIC_LINK is defined

james4k:
Avoids the need to specialize library loading code for new platforms that only support static linking.

### DIFF
--- a/src/hx/Lib.cpp
+++ b/src/hx/Lib.cpp
@@ -207,7 +207,7 @@ typedef void (*SetLoaderProcFunc)(void *(*)(const char *));
 typedef void *(*GetNekoEntryFunc)();
 typedef void (*NekoEntryFunc)();
 
-String GetFileContents(String inFile)
+static String GetFileContents(String inFile)
 {
 #ifndef _WIN32
    FILE *file = fopen(inFile.__CStr(),"rb");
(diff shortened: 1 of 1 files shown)


james4k:
The failed check here appears to be fixed by https://github.com/HaxeFoundation/hxcpp/commit/53bdca01aebb257d9f2766980dec937c35171dde

hughsando:
STATIC_LINK means we are creating a library - it does not necessarily mean that we will not be using dynamic loading.  I think something like HXCPP_NO_DYNAMIC_LOADING which can be set on the command line, or in hxcpp.h from the platform might be the way to go.

james4k:
OK, I've taken a closer look and I see dynamic loading would still be used, but at the same time there is some code that conflicts with this. If I'm not mistaken, __loadprim here does not do any dynamic loading:
https://github.com/HaxeFoundation/hxcpp/blob/4cf1165a48b54f6b139ae42012fed74be0638e8f/src/hx/Lib.cpp#L430

Not sure I understand, but is this an exception? What kind of dynamic loading is expected in the STATIC_LINK case?

hughsando:
The case I see is something like building a "libMyEngine.a" from hxcpp to link into external code using static_linl.  This library may, in turn, load something like NME dynamically.
You are right though, that code looks like it does not work with STATIC_LINK defined.  Not 100% sure why its disabled for SCRIPTABLE either.
The logic is a bit messy, but I think the key requirements are 1. Some platforms simply do not allow dynamic linking - this should apply to getProcAddress and _loadPrim.  2. Some developers may  choose to disable dynamic loading for reasons of their own.
I'm tempted to move this logic into the xml files.

james4k:
Some of the changes made since I created this PR have gotten part of the way there already. Will submit a separate PR that tightens up includes a bit and consolidates some of the preprocessor conditionals, without changing the logic.

PR #376, opened 2016-02-17
WinRT fixes: threads, 64bit, static compile, optional MD abi

madrazo:
Removed some WinRT code since now it is available in Windows 10 UWP and
previusly unavailable in Windows Store 8. Using 

    (_WIN32_WINNT == _WIN32_WINNT_WIN8) 

in case that someone is still interested in compiling for WinRT for Windows 8, but requires to fix toolchain.  

Encapsulated code for store-api with   

    defined(__cplusplus_winrt) 

so it can compile with the  -MD abi (-DABI="-MD" or -DHXCPP_WINRT_NATIVE_LIBS flags)

In this way, you can compile your
static libs for winrt more similar to C++ and only need to use the store-api in a Main.cpp file compiled with the ZW abi. Reference: https://msdn.microsoft.com/en-us/library/mt186162.aspx

### DIFF
--- a/src/hx/Thread.cpp
+++ b/src/hx/Thread.cpp
@@ -3,7 +3,7 @@
 #include <hx/Thread.h>
 #include <time.h>
 
-#if defined(WINRT) && (_WIN32_WINNT == _WIN32_WINNT_WIN8)
+#if defined(HX_WINRT) && (_WIN32_WINNT == _WIN32_WINNT_WIN8)
 using namespace Windows::Foundation;
 using namespace Windows::System::Threading;
 #endif
@@ -281,7 +281,7 @@ Dynamic __hxcpp_thread_create(Dynamic inStart)
 	hx::GCPrepareMultiThreaded();
 	hx::EnterGCFreeZone();
 
-   #if defined(WINRT) && (_WIN32_WINNT == _WIN32_WINNT_WIN8)
+   #if defined(HX_WINRT) && (_WIN32_WINNT == _WIN32_WINNT_WIN8)
 
    bool ok = true;
    try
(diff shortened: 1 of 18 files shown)


madrazo:
WinRT code for Win8 could be removed but I have leaved it in case that someone is interested in fixing its tool chain...

madrazo:
I use a a Main.cpp as a Haxe "template" (as in openfl), but I need to output "::" since WinRT uses a lot of namespaces. I workaround this using "::_::" to output "::" when parsing in  https://github.com/HaxeFoundation/haxe/pull/4833 but it there may be a better notation. For example writing an empty string "::::" instead of "::_::"

src/hx/Thread.cpp
```diff
@@ -3,7 +3,7 @@
 #include <hx/Thread.h>
 #include <time.h>
 
-#ifdef HX_WINRT
+#if defined(HX_WINRT) && (_WIN32_WINNT == _WIN32_WINNT_WIN8)
```
james4k: Should this be checking for `__cplusplus_winrt` instread of `_WIN32_WINNT_WIN8`? Maybe I misunderstand, but these are WinRT APIs, so building with the WinRT runtime would be necessary.

Also the same for similar changes in the rest of the PR.

james4k:
Would it make sense to get rid of the `HXCPP_WINRT_NATIVE_LIBS` option and make that behavior the default?

madrazo:
Windows 8.x RT would require the winrt api for threads, so it would be better just to deprecate that code and make HXCPP_WINRT_NATIVE_LIBS the default for static linking.  In Windows 10 RT you don't need the code inside _WIN32_WINNT_WIN8 since you can use the c++ thread API.

madrazo:
Favored #391

PR #389, opened 2016-03-02
improve portability of runtime with game consoles

james4k:
These changes address some problems with various proprietary SDKs used for game consoles. Originally, I wanted to have a number of smaller PRs each addressing separate platforms, but there is a lot of overlap in many cases. 

Most of these changes involve being more specific when encountering non-POSIX systems, quirks in POSIX support when it comes to date/time APIs, and specializing certain uses of Windows APIs as desktop-only.

This is not a holistic set of changes. Anything involving actual use of platform SDKs, even to create a thread or allocate memory, has been excluded due to NDAs. How this is currently approached in OpenFL's Lime framework is by moving relevant system calls in hxcpp's runtime into a common API. This API is implemented only in header files, that can then be replaced by a user-provided set of header files when certain macros are defined. Not super keen on this approach, but not sure of a better alternative. Will be submitting a PR soon that can take specific feedback.

[TODO: link to the above mentioned PR]

### DIFF
--- a/include/hx/Operators.h
+++ b/include/hx/Operators.h
@@ -117,7 +117,7 @@ inline L& UShrEq(L &inLHS, R inRHS) { inLHS = hx::UShr(inLHS,inRHS); return inLH
 template<typename L, typename R>
 inline L& ModEq(L &inLHS, R inRHS) { inLHS = DoubleMod(inLHS,inRHS); return inLHS; }
 
-#ifdef __GNUC__
+#if defined(__GNUC__) || defined(__SNC__)
 template<typename R>
 inline hx::FieldRef AddEq(hx::FieldRef inLHS, R inRHS) { inLHS = inLHS + inRHS; return inLHS; }
 template<typename R>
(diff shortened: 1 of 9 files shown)


madrazo:
Hi @james4k 
Please take a look to
https://github.com/HaxeFoundation/hxcpp/pull/391/files
I tried to merge your commit with mine but haven't try it.
Hopefully It would make easier for @hughsando to resolve the conflicts
Cheers!

james4k:
Hmm, OK. Maybe I will remove my Windows related changes.

With your changes I can probably have HX_WINRT defined for the Xbox One XDK which I currently do not do, largely due to the native vs WinRT dichotomy which you seem to address.

Will update you when I know more, @madrazo.

madrazo:
For Windows 10, the APIs have become similar but there are some differences. For example, you can define HXCPP_CAPTURE_x64 in GcRegCapture.h for XB but not for WinRT. In the DXTK, they use the 

    if defined(_XBOX_ONE)

So maybe it is OK to use HX_WINRT for both targets and use _XBOX_ONE when there is a specialization such as this one.

Also check out that you can compile static libs with MD abi (not using the store api). The idea is that only the main.cpp uses -ZW
 I added HXCPP_WINRT_NATIVE_LIBS flag but could be done with ABI=-MD
(You can know if using -ZW in code using the __cplusplus_winrt define)

james4k:
Yeah, I went the direction I went because it seemed that HX_WINRT implied the use of the WinRT runtime (`-ZW`), so I decided that would not make sense. But with your changes, it appears the meaning is shifting to mean the new WINAPI families in general, and not necessarily using the WinRT APIs or runtime.

james4k:
These changes were merged with related WinRT changes from @madrazo https://github.com/HaxeFoundation/hxcpp/pull/391

PR #391, opened 2016-03-04
Winrt and console

madrazo:
Merge of PRs #389 (and #376)

### DIFF
--- a/project/libs/std/Sys.cpp
+++ b/project/libs/std/Sys.cpp
@@ -15,7 +15,7 @@ int __sys_prims() { return 0; }
 #	include <windows.h>
 #	include <direct.h>
 #	include <conio.h>
-#   include <locale.h>
+#	include <locale.h>
 #else
 #	include <errno.h>
 #ifndef EPPC
@@ -46,9 +46,8 @@ int __sys_prims() { return 0; }
 #endif
 #endif
 
-#ifdef HX_WINRT
+#if defined(HX_WINRT) && !defined(_XBOX_ONE)
 #include <string>
-#define WINRT_LOG(fmt, ...) {char buf[1024];sprintf(buf,"****LOG: %s(%d): %s \n    [" fmt "]\n",__FILE__,__LINE__,__FUNCTION__, __VA_ARGS__);OutputDebugString(buf);}
 #endif
 
 #ifndef CLK_TCK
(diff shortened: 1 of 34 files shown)


madrazo:
We could get rid of the *ugly _WIN32_WINNT == _WIN32_WINNT_WIN8 (deprecate Windows 8.x store apps)

madrazo:
Seems to work OK for Windows target

madrazo:
I am thinking on closing this and resend fewer commits.

project/libs/std/Sys.cpp
```diff
@@ -47,8 +47,8 @@ int __sys_prims() { return 0; }
 #endif
 
 #ifdef HX_WINRT
-#include <hx/Thread.h>
-#include <hx/Tls.h>
+#include <string>
```
james4k: Did you mean `string.h`?

james4k:
I am testing this PR, and have a few cleanup changes I want to get in. Are you going to start a new PR, or should I base them on this branch?

Also, not exactly related to this PR, but should this line be in the toolchain XML?

https://github.com/HaxeFoundation/hxcpp/blob/707dfbf0b16015eebd042dbd81b0e323e9d10f87/toolchain/msvc-toolchain.xml#L50

This should be defined by the platform, right?

madrazo:
I think it is OK to use this PR, and if @hughsando request it, I can make a new one

Tried it without the WINAPI_FAMILY_APP flag and got an error in Math.cpp
 error C2039: 'RegisterClassA': is not a member of 'hx'

I think it is OK to define it:
https://msdn.microsoft.com/en-us/library/windows/apps/hh700130.aspx

Maybe it can be automatically defined in some way but I cannot find the way.

I added <string> for using wstring 
https://msdn.microsoft.com/en-us/library/wt3s3k55.aspx

madrazo:
Removed use of
< string > 
for _XBOX_ONE

james4k:
https://github.com/madrazo/hxcpp/compare/Winrt-and-console...james4k:winrt-and-console

james4k:
Is this use of WinRT APIs necessary when dynamically linking? If we could clean out all WinRT API usage in the hxcpp runtime/stdlibs, that might be ideal.

madrazo:
I think that you require store api  https://msdn.microsoft.com/en-us/library/mt186162.aspx (Using a Win32 DLL in a Universal Windows Platform App)

I believe that I am finish with this now...  do you need me to add that code in Lib.cpp ?

james4k:
I see, OK. Thanks.

Yes, the Lib.cpp change fixes a problem where the order of the conditionals means that the stubs are not used for STATIC_LINK if _WIN32 is defined.

james4k:
OK, all looks good from here.

madrazo:
Ups. Tried to "squash" the commits but didn't go well. I am closing this and resending a new PR.

PR #393, opened 2016-03-10
SSL lib

pperidont:
SSL bindings for neko, using mbedTLS ( https://tls.mbed.org/ )

The corresponding neko lib PR is here : https://github.com/HaxeFoundation/neko/pull/114

Main features :
- SSL sockets for client and server :
 - standard certificate and commonName verification
 - use certificate / key
  - Server Name Indication
- Load Certificate from file or directory
- Load Private / Public Key
- Digest signature and verification using public / private key

The corresponding Haxe wrappers are available at https://github.com/motion-twin/haxe/tree/ssl

Some tests and examples are available at : https://github.com/pperidont/haxe-ssl-tests


ncannasse:
I have merged the neko version already.
@hughsando I think it would be great to have SSL support shipping by default with hxcpp as well, so people don't have to install/rely on a 3rd party lib to make a single web service call.
This way we will have ssl socket as part of Haxe stdlib.

ncannasse:
@hughsando up for 3.3 please :)

thomasuster:
@pperidont I use sys.net.Socket on neko server and iOS / android / mac clients. Will this allow me to use some kind of equivalent SSL socket? Or is there more work required for implementation / ios / android?

pperidont:
@thomasuster Yes, this lib will allow you to use ssl sockets with sys.ssl.Socket. 
You can find the haxe wrappers, that is very similar to sys.net.Socket here : https://github.com/motion-twin/haxe/blob/ssl/std/cpp/_std/sys/ssl/Socket.hx

It works on Linux, Windows, Mac, Android and iOS. It should work on others platforms, but it's not tested yet.
Loading of system default Certificates Authorities (CA)  is not supported on iOS.

thomasuster:
@pperidont Very cool, thank you for the PR! I noticed the link to a separate repo for testing, if your PR gets taken would you consider opening a PR to add tests to the haxe test suite, I'd love to give this source some coverage so haxe maintainers have a better chance of long term success.

Simn:
I'm gonna merge this tomorrow if no one else does.

I'll still blame other people for it if something goes wrong though!

PR #396, opened 2016-03-15
WinRT and Game Consoles

madrazo:
- james4k: changed order in Lib.cpp
- james4k: game console porting fixes in time functions

WinRT fixes:
- 64 bit compile
- Automatically set latest installed Windows 10 SDK paths using
environment variables (tested with both older and newest SDKs )
- Deprecates WinRT 8.x (cleaner code, threading was not working ok anyway)
- Encapsulates WinRT "store-api" with __cplusplus_winrt to enable
compiling of native static libs.
- Set to compile native libs (MD ABI) as default when static linking.
- Use GetTickCount64() for WInRT instead of clock() which should be more robust (this function is not
available on WindowsXP).
- fix __hxcpp_dbg_checkedThrow to work with winrt

- some style fixes

Merges from
https://github.com/HaxeFoundation/hxcpp/pull/391
https://github.com/HaxeFoundation/hxcpp/pull/389
https://github.com/HaxeFoundation/hxcpp/pull/376

### DIFF
--- a/include/Dynamic.h
+++ b/include/Dynamic.h
@@ -335,7 +335,7 @@ inline bool operator <= (bool inLHS,const Dynamic &inRHS) { return false; }
 inline bool operator >= (bool inLHS,const Dynamic &inRHS) { return false; }
 inline bool operator > (bool inLHS,const Dynamic &inRHS) { return false; }
 
-#ifdef HX_WINRT
+#if defined(HX_WINRT) && defined(__cplusplus_winrt)
 // Try to avoid the compiler using injected Box::operator int and Dynamic(null) when doing ==
 template<typename T>
 bool operator==(Platform::Box<T> ^inPtr, nullptr_t)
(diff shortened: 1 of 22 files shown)


hughsando:
Is this the one ready to go?

hughsando:
I have manually merged this one.

madrazo:
Thanks!

hughsando:
And thank you.
You should be aware that I have now restructured the stdlibs code in hxcpp.
Currently there are 2 locations for changes such as these - the files you have changed here will be used for hxcpp 3.2.x, and for hxcpp 3.3+ the changes should go in src/hx/libs/std/*cpp.
These files will built in as part of the runtime, rather than linked in at the end, so it should be easier to support a wider range of compilers without the need to pre-build libraries.
I did my best to replicate your changes into the 3.3+ files, but it may need some additional testing.

PR #399, opened 2016-03-23
Higher entropy seed for Math.random()

thomasuster:
I'm getting duplicate uuid from https://github.com/andyli/casahx, @andyli suggested looking into seed. Current implementation was putting a lot of emphasis on pid which is OS specific, and I'm betting may be less random than we think on android / iOS / mac. The implementation I found hear seems to be a lot more random. http://stackoverflow.com/questions/8920411/possible-sources-for-random-number-seeds

Testing:
* Pakka Pets runs fine
* Test suite passes
* Tried printing 10 random numbers spawning multiple processes and skimmed the output to ensure no duplicates

What do you think?

### DIFF
--- a/src/Math.cpp
+++ b/src/Math.cpp
@@ -152,20 +152,23 @@ void Math_obj::__boot()
 
 	unsigned int t;
 #ifdef HX_WINDOWS
+    unsigned int t;
 	t = clock();
    #ifdef HX_WINRT
 	int pid = Windows::Security::Cryptography::CryptographicBuffer::GenerateRandomNumber();
    #else
 	int pid = _getpid();
    #endif
+   srand(t ^ (pid | (pid << 16)));
 #else
-	int pid = getpid();
-	struct timeval tv;
-	gettimeofday(&tv,0);
-	t = tv.tv_sec * 1000000 + tv.tv_usec;
-#endif	
+   long s, seed, pid;
 
-  srand(t ^ (pid | (pid << 16)));
+   pid = getpid();
+   s = time(0);
+
+   seed = std::abs(((s*181)*((pid-83)*359))%104729);
+   srand(seed);
+#endif
   rand();
 }
 


thomasuster:
I'm thinking maybe this isn't enough. Maybe this with nano seconds?

hughsando:
If you use modulo 104729, then there will only be 104729 different random number sequences.
Combining the time into the seed seems like a very good idea, otherwise someone could "guess the next card" based on the pid.
I think if you are after a guid, you need to combine these sources of entropy in a different way - not  by seeding a random number, which compresses the number of significant bits.
Eg, build a string with pid + time in ns + phone model + user id of process + random() +  whatever you can think of.  Then md5 this string.  If you then use this as a seed, you will be back to the situation of the number of bits in the seed.

ncannasse:
If you want to some quite solid hashing of Int32, you can use https://github.com/ncannasse/heaps/blob/master/hxd/Rand.hx#L24

hughsando:
For a gui, I think you want more than 32 bits, otherwise you would expect
clashes with 100k random entries.
In fact, depending on how you choose your seed you could get a very high
chance of clash (eg, pid)

thomasuster:
@hughsando Ahh, good point. You're saying if we did it perfectly with math.rand then maybe we hit a chance of a clash after about 2^32 uuids, about 2,147,483,647, where as UUID should be one chance of clash in about 3.4^38. I'm going to work on my own UUI imp then in a new lib.

hughsando:
Yes 32 bits = 64k * 64k.  So say your first 64k entries are unique.  The chance of a random new one clashing is 1/64k - not too bad - until you do this 64k more times and the chances get closer to 1.
64 bits and the odds look way better, 128 bits is the standard I think.

andyli:
Umm... I wonder how libuuid implement it. [Its manpage](http://linux.die.net/man/3/libuuid) states that

> The UUIDs generated by this library can be reasonably expected to be unique within a system, and unique across all systems.

So it must not be a naive implementation similar to casalib's...

thomasuster:
So I realized I could count the total number of users Pakka Pets has, this would likely give an indication of how many UUIDS could be generated, and therefore approximate how good the current seeding of Math.rand is. The number is 117,606.

thomasuster:
I tried an all haxe approach, but could only get about 48 bits worth of randomness for the seed, the attempt is here if you want to see seeds I thought of. https://github.com/thomasuster/uuid/blob/master/src/thomasuster/UUIDFactory.hx

I ended up making an extension for getting 32 random secure bits, and then writing a UUID factory inside the pakka library using 4 calls to getRandomSecureBytes, then md5, then toUpper case and split with some dashes.

iOS: https://github.com/thomasuster/extension-secureRandom/blob/master/project/iPhone/SecureRandom.mm
android: https://github.com/thomasuster/extension-secureRandom/blob/master/android/src/com/thomasuster/SecureRandom.java

Issue #404, opened 2016-04-03
Comparison array and null


ibilon:
On haxe 3.2.1 the following code:
```
class Test {
	public static function main () 	{
		test([1,2,3,4]);
		test();
	}	
	public static function test (array:Array<Int>=null) 	{
		if (array == null)
			trace([]);
		else
			trace(array);
	}
}
```
pass haxe compilation but fails with g++, something about impossible comparison between array and null.
output: http://pastebin.com/k66RE4Sm

On latest haxe dev I get:
```
Error: /home/ibilon/Code/Haxelibs/hxcpp/src/hx/StdLibs.cpp: In function ‘const char* hx::Init()’:
/home/ibilon/Code/Haxelibs/hxcpp/src/hx/StdLibs.cpp:160:14: error: found ‘:’ in nested-name-specifier, expected ‘::’
       catch(e:Dynamic)
              ^
/home/ibilon/Code/Haxelibs/hxcpp/src/hx/StdLibs.cpp:160:13: error: expected type-specifier before ‘e’
       catch(e:Dynamic)
             ^
/home/ibilon/Code/Haxelibs/hxcpp/src/hx/StdLibs.cpp:160:14: error: expected ‘)’ before ‘::’ token
       catch(e:Dynamic)
              ^
/home/ibilon/Code/Haxelibs/hxcpp/src/hx/StdLibs.cpp:160:14: error: expected ‘{’ before ‘::’ token
/home/ibilon/Code/Haxelibs/hxcpp/src/hx/StdLibs.cpp:160:22: error: expected unqualified-id before ‘)’ token
       catch(e:Dynamic)
                      ^
```

ncannasse:
Fixed by #405

ibilon:
It fixes the second part but it still fails on haxe 3.2.1 with things like:
```
Error: ./src/haxe/Resource.cpp: In static member function ‘static haxe::io::Bytes haxe::Resource_obj::getBytes(String)’:
./src/haxe/Resource.cpp:35:20: error: no match for ‘operator==’ (operand types are ‘Array<unsigned char>’ and ‘null’)
  bool tmp = (array == null());  HX_STACK_VAR(tmp,"tmp");
                    ^
./src/haxe/Resource.cpp:35:20: note: candidates are:
In file included from /home/ibilon/Code/Haxelibs/hxcpp/include/hxcpp.h:319:0,
                 from ./src/haxe/Resource.cpp:1:
[...]

PR #405, opened 2016-04-03
Fix haxe syntax -> cpp syntax

ibilon:
You can see a compiler output of this in #404


ncannasse:
Seems legit.

hughsando:
he he.  Not sure how that survived testing. Too many languages is barely enough.

PR #408, opened 2016-04-04
Convert SSL to native externs

pperidont:
Follow https://github.com/HaxeFoundation/hxcpp/pull/393 and cef6946c40df526dcccfcda30b8170c4d31ea774

I've done the conversion of the SSL lib to the new native extern system.

I think SSL should be in Haxe standard lib, because it's becoming an essential complement to http : all the most common web API already require https.

A lib on haxelib can do it for hxcpp, but for neko it will not be very convenient (for binaries distribution).

And this way we can have a common sys.ssl.* API in the Haxe standard library (as proposed in the Haxe PR https://github.com/HaxeFoundation/haxe/pull/5014 )

The library sync process should be fairly simple : the files in thirdparty/mbedtls come directly from mbedtls release with only very small modifications :
 - some small changes in config.h (this file has a lot of comments, but it's a simple list of defines) 
 - an additional file threading_alt.h (a typedef used on windows only)


hughsando:
Thanks for doing this work -  I know it is not fun.
Lets move forward assuming it will be in "core" functionality.
I have not looked too closely, but I think you may need some Enter/Exit GCFreeZone calls around calls that could potentially block for a reasonable amount of time.  This ensures that if a different thread wants to do a garbage collection, then they do not need to wait the the socket call to return.
I will accept this for now, and have a closer look later.

PR #412, opened 2016-04-06
BuildTool: don't hardcode verbose output to travis

Gama11:
Pretty sure this is the reason why Flixel's cpp dev builds are still verbose even though I explicitly define `HXCPP_QUIET`.

I general it seems like bad practice to hardcode this to a particular CI system?


Gama11:
Guess this was probably leftover from debugging: https://github.com/HaxeFoundation/hxcpp/commit/2fc570f9c68e5c057a5689ac30fc065a98b449d8

hughsando:
This was when I actually wanted more info from travis.  Have not needed it for a while.

PR #414, opened 2016-04-08
New "overwrite" option in CopyFile.

aW4KeNiNG:
Hi.

I have a large list of "copyFile"s in my build. These files don't change over time and it is really slow when i try to compile the application. So i have added a new "overwrite" option, with "true" by default.

### DIFF
--- a/tools/hxcpp/BuildTool.hx
+++ b/tools/hxcpp/BuildTool.hx
@@ -1,3 +1,4 @@
+import CopyFile.Overwrite;
 import haxe.io.Path;
 import haxe.xml.Fast;
 import haxe.Json;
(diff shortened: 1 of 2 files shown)


hughsando:
It might be nice to have a system that checks time stamps (isOutOfDate).  This could probably be on by default, with maybe "alwaysOverwrite" to force overwriting, not mater what the stamp.

aW4KeNiNG:
Maybe the "overwrite" option could be a list of options instead of bool. For example, it can be "always", "never", "only_old_date"... Or something like that. What do you think?

hughsando:
Sound good = "always", "never", "ifNewer"   Currently, the implementation
is "always", so I guess this should be the default to avoid causing
problems.

aW4KeNiNG:
Done :)

PR #426, opened 2016-04-18
Fix getNdkVersion

kevinresol:
see: http://try.haxe.org/#5601e

### DIFF
--- a/tools/hxcpp/Setup.hx
+++ b/tools/hxcpp/Setup.hx
@@ -54,7 +54,7 @@ class Setup
 
    static public function getNdkVersion(inDirName:String):Int
    {
-      var extract_version = ~/(android-ndk-)?r(\d+)*/;
+      var extract_version = ~/\/(android-ndk-)?r(\d+)*/;
       if (extract_version.match(inDirName))
       {
          return Std.parseInt( extract_version.matched(2) );


hughsando:
I think that this should probably be "^" (anchor start) like in the other regex, instead of slash.  Does this work for you?

kevinresol:
Without my fix, it doesn't work, because `inDirName` is a full path like `/usr/somewhere/android-ndk/r10b`. So I think a `^` will not work either...

hughsando:
I see, it is using the full path.
Then I think your solution is good, although I would be tempted to add ".*" and the beginning to greedily consume other possible matches and leave just the last.

kevinresol:
I am not very good at regex, how about `$` at the end instead?
e.g. `~/\/(android-ndk-)?r(\d+)*\/?$/;`

hughsando:
I have just put something like this in.  I think the main problem is the star at the end - this allowed it to match against a simple 'r', which is obviously not good.  The slash provides a good anchor as well.

hughsando:
Maybe $ as end would have been good too.  Maybe as well?

hughsando:
Need room for the letter at the end - not sure what the range could be here. [a-z]? maybe

hughsando:
I hope this is working now, so closing.

PR #429, opened 2016-04-19
[WinRT] Fix Sys.print and trace

madrazo:
Needed a fix in the WINRT_PRINTF macro ("fmt" -> fmt) and implemented in __trace function.


madrazo:
Also, snprintf should be safer. Thanks.

PR #430, opened 2016-04-21
consolidate preprocessor logic for lib loading

james4k:
Also tightens up includes for improved portability.


james4k:
A worthwhile step after this might be to simplify the conditionals, but this should exactly match the logic from before this PR.

PR #431, opened 2016-04-24
overridable system APIs

james4k:
This is the last major set of changes needed to support building Haxe C++ applications with proprietary, closed SDKs. These changes move most of the system calls that hxcpp's runtime makes into header files that the user can override with their own by defining a macro.

Not sure how I feel about this approach, though it gets the job done. Open to feedback, big and small.

### DIFF
--- a/include/hx/Memory.h
+++ b/include/hx/Memory.h
@@ -1,0 +1,24 @@
+#ifdef HX_MEMORY_H_OVERRIDE
+// Users can define their own header to use here, but there is no API
+// compatibility gaurantee for future changes.
+#include HX_MEMORY_H_OVERRIDE
+#else
+
+#ifndef HX_MEMORY_H
+#define HX_MEMORY_H
+
+inline void *HxAlloc(size_t size) {
+   return malloc(size);
+}
+
+inline void *HxAllocGCBlock(size_t size) {
+   return HxAlloc(size);
+}
+
+inline void HxFree(void *p) {
+   return free(p);
+}
+
+#endif
+
+#endif
--- a/include/hx/Thread.h
+++ b/include/hx/Thread.h
@@ -117,9 +117,9 @@ inline int HxAtomicDec(volatile int *ioWhere)
 #if defined(HX_WINDOWS)
 
 
-struct MyMutex
+struct HxMutex
 {
-   MyMutex()
+   HxMutex()
    {
       mValid = true;
       #ifdef HX_WINRT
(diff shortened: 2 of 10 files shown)


hughsando:
This seems pretty sensible.  The "My..." names were a mistake.
Out of interest, how do you inject the alternate versions from different compilers?

madrazo:
"CreateThread" should work for HX_WINRT (Windows 10) even with "__cplusplus_winrt".

james4k:
OK, yeah I should explain the other side of it which is a little more strange.

In the toolchain XML, I specify a couple of extra flags for a "forced include" and an include path for the overriding headers, and then rely on the fact that HXCPP_API_LEVEL is only >0 if building Haxe code.

The "forced include" contains:
```cpp
// Bit of a hack, but HXCPP_API_LEVEL is greater than zero only on haxe
// generated code, and hxcpp.
#if HXCPP_API_LEVEL > 0

#define HX_MEMORY_H_OVERRIDE <hx_memory_override.h>
#define HX_THREAD_H_OVERRIDE <hx_thread_override.h>
#define HX_TLS_H_OVERRIDE <hx_tls_override.h>

#endif
```

james4k:
@madrazo Ah, nice. I'll submit a PR to fix that soon.

hughsando:
Ok, though it was something like that.
One thing we could do, would be to "#include <hx/ToolchainConfig.h>" or
similar, and have a default implementation that either does nothing, or we
put all the built-in sys functions in there.
Then you can use an -Iinclude-path to override this file with your specific
version of this.
Might be neater than a force-include, which might prevent pre-compiled
headers

james4k:
Something along those lines would work well. I would like to avoid relying on the search order of the include paths, though. Some kind of option in the toolchain.xml, or an option/define passed through the CLI, that would then have hxcpp include such a file, and then can safely let it fail if no such header was found.

hughsando:
Yes, seems safer, like
```
#ifdef HXCPP_CUSTOM_CONFIG
#include <HxcppCustomConfig.h>
#endif
```
Another way would be to put them in somewhere that is not normally in the include path ($HXCPP/include/config/windows/HxcppSys.h) then ensure that every compiler explicitly adds one config path to the include path.

james4k:
Either of those sound good to me. Having each toolchain specify a config path is probably more consistent, and if the header contains the sys funcs themselves, maybe that is a simpler design. Could see windows/posix config headers. Although, that means all Haxe C++ would include this code.

PR #445, opened 2016-05-07
Support Unicode properties in regexp

minoki:
Currently, support for Unicode properties in PCRE is not enabled and `\p` patterns cannot be used (See #367).

Since it can be enabled with a flag (`-DSUPPORT_UCP`), why not do so?


hughsando:
Looks simple enough!

PR #451, opened 2016-05-12
Do not use boolean and operator when converting to Int64 / UInt64

waneck:
Without this, the following behaviour happens:
```haxe
class Main {

  static function main() {
    var i:cpp.Int64 = getDynamic();
    trace(i); // 1 :(
  }

  static function getDynamic():Dynamic {
    var i:cpp.Int64 = 42;
    return i;
  }

  }
```

That was a hard one to catch :P


waneck:
Travis just segfaulted on this build, but it seems to be a Haxe issue - it segfaulted on `nekotools boot`

hughsando:
Thanks - at least you did not test with "1"

andyli:
> it segfaulted on `nekotools boot`

Arghh, it must be related to the recent changes I made, looking at it now.

PR #466, opened 2016-06-07
Added column number break functionality to hxcpp, hxcpp-debugger

tekgek:
* Added column number fields to code-position-related function
* Added a nextLine function so that developers can step over an entire line of code
* Added column number check to FindFileLineBreakpoint()


hughsando:
Sorry I have not commented on this earlier.
Is this something you are still interested in?

EricBishton:
Yes, Hugh!  Very, very, much.  There are accompanying changes to the compiler, but they are fairly small.  Simon was waiting on those changes until he got some feedback from you on these changes.  

TiVo spent a good bit of time and money getting these changes created by the UCSC team, and they allow the debugger to set breakpoints on and step between individual expressions on a single line.  It's a much more granular debugging workflow.

hughsando:
I have some mixed feelings about this one.
I'm sure you are aware that there is not a 1:1 relationship between haxe and cpp code - especially at the column level.  So understand better, are you looking to be able to break at the "HXDLIN" macros, rather than just the "HXLINE" macros, which only occur then the line changes value?

My main concerns are that old(ish) haxe need to work with new hxcpp, and perhaps adding complexity/speed when it is only sometimes needed.  Generally hxcpp can work with multiple versions by either conditionally defining the macos, or adding new macros with different names, and adding default args to non-macro functions.

Perhaps these concerns can be overcome by a haxe define, like -D debug_columns.  Turning this on would generate the extra info in the output .cpp file, and add the HXCPP_DEBUG_COLUMNS define to the compile phase.  The debugger library could then also use the debug_columns condition to work out what commands are available.

PR #511, opened 2016-09-14
Fix DEFINE_PRIME5v and DEFINE_PRIME7v preprocessor typos

Tw1ddle:
Those preprocessor macros look like they have typos.

Attempts to send some arrays of strings from Haxe to C++ using CFFI PRIME took me here. Is there any documentation or example that could help with that? Thanks :-)


Tw1ddle:
Ah, I found this post which explains that for "complex" types you should continue with the old CFFI way. I'll do as it says: http://gamehaxe.com/2016/02/10/cffi-prime-for-cffi-users/ :smile:

hughsando:
Had the right number of 'v's, just on the wrong places!

PR #518, opened 2016-10-13
Avoid null pointer exception for a

derRaab:
I ran into an issue where a is null and that kills the application. So I figured it might be a good idea to avoid that. Fun fact: if (0==a) doesn't work. Any ideas why that is?


derRaab:
Does anyone know why one testsuite doesn't start?

Simn:
Travis OS X has been having some issues, but that's unrelated to your pull request.

PR #530, opened 2016-11-08
Fix ndk 13b compile

madrazo:
Include ".x" directory


madrazo:
Sorry. This is incorrect.

Issue #557, opened 2017-01-31
iOS: crashing when Bytes getFloat from unaligned data on "Release" build


madrazo:
I though that [this](https://github.com/HaxeFoundation/hxcpp/issues/522) was solved, but when setting the "Release" profile (sets -Os as default), it crashes here (tested on iPod touch 5th -32 bit CPU-)
[https://github.com/HaxeFoundation/hxcpp/blob/master/include/hx/StdLibs.h#L306](https://github.com/HaxeFoundation/hxcpp/blob/master/include/hx/StdLibs.h#L306)
with EXC_ARM_DA_ALIGN

A work around is to change
`*dest = *src;`
to
`memcpy((void*)dest,src,sizeof(float));`
or, byte by byte as before

PR #561, opened 2017-02-02
iOS fix getFloat from unaligned data on Release

madrazo:
Prevent EXC_ARM_DA_ALIGN exception on iOS Release when reading a unaligned float. 
https://github.com/HaxeFoundation/hxcpp/issues/557

TODO: check __hxcpp_align_set_float32, __hxcpp_align_set_float64 and __hxcpp_align_get_float64

### DIFF
--- a/include/hx/StdLibs.h
+++ b/include/hx/StdLibs.h
@@ -301,9 +301,13 @@ inline float __hxcpp_align_get_float32( unsigned char *base, int addr)
    if (addr & 3)
    {
       float buf;
+      #ifdef IPHONE
+      memcpy((void*)&buf,(base+addr),sizeof(float));
+      #else
       unsigned int *dest = (unsigned int *)&buf;
       const unsigned int *src = (const unsigned int *)(base+addr);
       *dest = *src;
+      #endif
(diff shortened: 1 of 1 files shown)


hughsando:
So does it appear you can't do unaligned int moves on IOS release mode?  I should check the disassembly.
If this is the case, then the int-read code may need checking too.
I wonder if this is a compiler/optimization flag that can be changed?

madrazo:
I couldn't find the flag but in this reference gives malloc and __packed as solutions
https://brewx.qualcomm.com/bws/content/gi/common/appseng/en/knowledgebase/docs/kb95.html
here nore abut malloc
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.faqs/ka3934.html

hughsando:
I decided to always do byte-by-byte copies in HXCPP_ALIGN_FLOAT for the 4 cases. I could be "overoptimizing" the int move into a float move, and I think the more recent arm processors are more finicky about alignment, and this case is to that common, so best be safe.

hughsando:
But thanks for your research.

PR #586, opened 2017-04-03
Fix Int64 to Int/Float casts

waneck:
Ever since https://github.com/HaxeFoundation/hxcpp/commit/5f29028d6063b37ff7915cff1b3b83b6beee103d#diff-24091b03a703274be4a56a48d140ed24R144 was done, any casts of Int64's to either Int or Float started to fail with a bad cast. This PR fixes this.

PR #588, opened 2017-04-04
Attempt at fixing #587

### LINKED ISSUE #587: Cppia failure when host-derived types implement a cppia interface
The code at https://gist.github.com/waneck/5dce5a14064a62b6be4b33905ccd3720 segfaults on latest hxcpp.

As far as I could understand, the issue is that interface slots are negative, which work correctly when a type was fully created by cppia. However, if that type derives from the host (like in this example, Test extends Main), this negative vtable slot is non-existent, which makes it access an unbound array position.

waneck:
Call `__GetScriptVTable()` if this is an interface call

Closes #587

This actually fixes it, but I think there must be a prettier (read: more correct :P ) fix

### DIFF
--- a/src/hx/cppia/Cppia.cpp
+++ b/src/hx/cppia/Cppia.cpp
@@ -2046,7 +2046,7 @@ struct CallMemberVTable : public CppiaExpr
    #define CALL_VTABLE_SETUP \
       hx::Object *thisVal = thisExpr ? thisExpr->runObject(ctx) : ctx->getThis(); \
       CPPIA_CHECK(thisVal); \
-      ScriptCallable **vtable = *(ScriptCallable ***)((char *)thisVal +scriptVTableOffset); \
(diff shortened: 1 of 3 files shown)


hughsando:
Actually, your fix looks ok.  In the case of host-implemented interfaces, the virtual call must be used to get the vtable, since the type is not known to be the interface delegate.

waneck:
I've pushed to the HaxeFoundaiton/hxcpp project by mistake, and quickly deleted it - this made both travis and appveyor fail. There's another travis/appveyor instance which is still running. I think the same issue happened on https://github.com/HaxeFoundation/hxcpp/pull/586 - which shows that both tests passed, but the `travis-ci/push` and `appveyor/branch` failed (because there was no branch by the time the tests were run)

hughsando:
I opened a bit of a can of worms when I tried to write some hairier tests for this.  I have incorporated this change in a larger fix, so thanks.
https://github.com/HaxeFoundation/hxcpp/commit/72187839359e250636c8518457acfb05d0adb402

waneck:
Thank you!

PR #591, opened 2017-04-26
Fixed for very big namespace, just added one more.

SergeyLabutin:


hughsando:
That's a big namespace!

PR #601, opened 2017-06-20
Add exit hooks, and make sure mspdbsrv is closed when hxcpp exits

### LINKED ISSUE #600: Compiler doesn't close stdout when compiling debug on Windows 64-bit on MSVC 2015
Okay, that's a somewhat specific issue, but it's very hard to understand what's going on when it happens

Prerequisites:
 * MSVC 2015
 * Make sure that mspdbsrv.exe is not running already on the background (otherwise it works)
 * Compile any hxcpp program with `-D HXCPP_M64 -debug`

What happens:
 * Hxcpp opens mspdbsrv.exe but doesn't close it. For some reason, this makes windows not close hxcpp's stdout. Other programs might then hang indefinitely waiting for the hxcpp compilation to finish

Sample code:

`Runner.hx`
```haxe
 class Runner {
   static function main() {
     var args = Sys.args();
     var cmd = new sys.io.Process(args.shift(), args);
     var stdout = cmd.stdout.readAll();
     trace(stdout);
     cmd.exitCode();
     trace('Ended');
   }
 }
```

`Main.hx`
```haxe
 class Main { 
     
   static function main() {
     trace("hello world");
   } 
   
 }
```

Run with `haxe --run Runner haxe -cpp cpp -D HXCPP_M64 -debug -main Main` . The program hangs indefinitely. You can then go on the task manager, and kill mspdbsrv manually to see it print stdout.
Note that this is not a problem with the sys.io.Process API - it happens with other unrelated programs that might call Haxe.

waneck:
Closes #600

Note that this PR is smaller than it looks - there were a lot of leading spaces that added up on the diff

### DIFF
--- a/tools/hxcpp/BuildTool.hx
+++ b/tools/hxcpp/BuildTool.hx
@@ -127,10 +127,10 @@ class BuildTool
 
 
       if (mDefines.exists("toolchain"))
-      {   
+      {
          if (!mDefines.exists("BINDIR"))
-         {   
-            mDefines.set("BINDIR", Path.withoutDirectory(Path.withoutExtension(mDefines.get("toolchain"))));  
+         {
+            mDefines.set("BINDIR", Path.withoutDirectory(Path.withoutExtension(mDefines.get("toolchain"))));
          }
       }
       else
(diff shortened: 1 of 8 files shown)


nadako:
Note that one can add `?w=1` to the URL with the diff and it'll ignore whitespace.

PR #604, opened 2017-07-02
warning: binding dereferenced null pointer?

thomasuster:
Hey Hugh

Anything that inlines the function I modified will spit out the following warning, might be mac only. It's just some that I've ignored for a long time, but this minor change seems to fix it. Does it look kosher to you? Not sure what the difference really is, but compiler seems to like it better.

Fix for hxcpp/include/null.h:132:61: warning: binding dereferenced null pointer to reference has undefined behavior [-Wnull-dereference]

### DIFF
--- a/include/null.h
+++ b/include/null.h
@@ -128,8 +128,11 @@ class null
 
      inline Dynamic &operator()(const AnyArg &a0=0, const AnyArg &a1=0, const AnyArg &a2=0,
             const AnyArg &a4=0, const AnyArg &a5=0, const AnyArg &a6=0,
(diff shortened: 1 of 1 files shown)


hughsando:
Yeah, the warning would get annoying. 
Hard to see what the difference between with what you have written.  The code is never going to get called, so I guess as long as the compiler is happy.
I think the 'best' fix would be to remove the "&" from the function, so it returns an instance.  Then your "nullDynamic" could just be "Dynamic nullDynamic; return nullDynamic".

thomasuster:
https://github.com/HaxeFoundation/hxcpp/pull/605

PR #605, opened 2017-07-02
Fixed binding dereferenced null pointer using template return.

thomasuster:


thomasuster:
Hey Hugh, this look good?

hughsando:
Great.  A pain for code that will not be called, but good to remove the warning.  Thanks.

PR #617, opened 2017-07-31
HXCPP tools speedup (WIP)

thomasuster:
Hey @hughsando 

This PR does a few things currently. Note this is NOT ready for merging yet. Still a work in progress.

* Multithread mCompiler.getCachedObjName
* Convert HXCPP build tool to CPP (for performance)
* A neko (-fallback) version is used while building the CPP version itself

But it seems to have problems with GC that I need your help on.
It keeps printing out
`CriticalGCError("Bad Allocation while collecting - from finalizer?");`
I've looked through the source files and I can't make sense of some of the locking mechanisms. It's also a 6,500 line file, it's very tough to figure out what's happening. I was able to print a stack trace where that CriticalGCError statement is. It looks to be triggering from LocalAllocator.callAlloc, which happens to be happening in another thread while a collect is occurring. What mutex or lock is suppose to stop this from happening?

```
0   BuildTool-debug                     0x000000010dcf898c _Z7handleri + 28
1   BuildTool-debug                     0x000000010dcfcf38 _ZN14LocalAllocator15PauseForCollectEv + 72
2   BuildTool-debug                     0x000000010dd06d99 _ZN14LocalAllocator9CallAllocEij + 73
3   BuildTool-debug                     0x000000010dcfd927 _ZN2hx11InternalNewEib + 407
4   BuildTool-debug                     0x000000010dd4c556 _ZN2hx9ArrayBaseC2Eiiib + 134
5   BuildTool-debug                     0x000000010dc29eca _ZN9Array_objIiEC2Eii + 90
6   BuildTool-debug                     0x000000010dc29e61 _ZN9Array_objIiEC1Eii + 33
7   BuildTool-debug                     0x000000010dc289cf _ZN9Array_objIiE5__newEii + 79
8   BuildTool-debug                     0x000000010dc2883b _ZN4haxe6crypto7Md5_obj8doEncodeE5ArrayIiE + 42651
9   BuildTool-debug                     0x000000010dc28bc7 _ZN4haxe6crypto7Md5_obj6encodeE6String + 151
```

It's also using this for now, https://github.com/thomasuster/haxe-threadpool, but I can refactor and/or duplicated it into HXCPP once all other problems are sorted. So let's not worry about that for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ test/extern-lib/gen-externs
 test/cffi/project/ndll
 test/snippets/messagebox/cpp
 hxcpp.n
+hxcpp-fallback.n
 .DS_Store
 
 *.swp
--- a/include/Array.h
+++ b/include/Array.h
@@ -348,9 +348,9 @@ public:
 
    mutable int length;
 
-   static inline int baseOffset() { return (int)offsetof(ArrayBase,mBase); }
-   static inline int allocOffset() { return (int)offsetof(ArrayBase,mAlloc); }
-   static inline int lengthOffset() { return (int)offsetof(ArrayBase,length); }
+   #pragma static inline int baseOffset() { return (int)offsetof(ArrayBase,mBase); }
+   #pragma static inline int allocOffset() { return (int)offsetof(ArrayBase,mAlloc); }
+   #pragma static inline int lengthOffset() { return (int)offsetof(ArrayBase,length); }
 
 protected:
    mutable int mAlloc;
--- a/tools/hxcpp/BuildTool.hx
+++ b/tools/hxcpp/BuildTool.hx
@@ -1,3 +1,5 @@
+import haxe.CallStack;
+import com.thomasuster.threadpool.ThreadPool;
 import CopyFile.Overwrite;
 import haxe.io.Path;
 import haxe.xml.Fast;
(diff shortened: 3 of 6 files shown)


hughsando:
The stack looks pretty legit - although I would expect to see some thread related functions a bit deeper.
It would be interesting to see where the thread that is doing the collect is at.  Presumably between where sgIsCollecting=true and sgIsCollecting=false.

The sequence should go like this:
One thread is calling Collect - you should be able to see this on one of the other threads. It should be doing stuff after: ` sgIsCollecting = true;`

To have got to this line,  all other threads should have passed the 'WaitForSafe' test - that is it should be in a gcfreezone, or mReadyForCollect should be set.

mReadyForCollect is set either once the thread is in the gcfreezone, or in pauseForCollect.
When it exits the gcfreezone, mReadyForCollect should be reset.
When it exits pauseForCollect, it calls mCollectDone.Wait - which should only get released from ReleaseFromSafe - but at this point sgIsCollecting should be false.

So the question is, how did the offending threads mReadyForCollect get set and not reset by the time it gets to the CallAlloc call?
If you printf-ed the setting/resetting of this variable, and the sgIsCollecting setting it may reveal the answer.
It might be a logic error in the transitions of gMultiThreadMode - setting this to true from the start would rule this out.

thomasuster:
@hughsando Thank you! Great information. I will look into those.

PR #620, opened 2017-08-03
Threaded get cached obj name

thomasuster:
This PR does two major things.
1) HXCPP build tool from neko to CPP
2) Multi-threads getCachedObjName

You must install this lib for this to build.
```
git clone https://github.com/thomasuster/haxe-threadpool.git
haxelib dev haxe-threadpool haxe-threadpool
```

To build
```
cd hxcpp/tools/hxcpp
haxe compile.hxml
```

To test try building some HXCPP project.

In my test project, this brings the HXCPP build time from 28 second down to 13 seconds!

The con is I occasionally get this in the output.

```
BuildTool(96053,0x70000ef86000) malloc: *** error for object 0x7faaf5d98000: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
```

I'm on OSX, so I'm not sure the best way to approach doing this. Xcode has an attach debugger feature, but it doesn't seem to pick up the process.


jcward:
Hmm, I tried to test this, but I don't think I successfully built it. I pulled your hxcpp and built using the instructions from the [main README](https://github.com/HaxeFoundation/hxcpp/blob/master/README.md), and set hxcpp dev to point to it, but it never complained that I haven't installed `haxe-threadpool`, and the build times were unchanged. So I don't think my rebuild of hxcpp was actually successful.

Aside: I've had this problem before, so I typically just rely on @hughsando's [build server](nmehost.com/hxcpp/) for nightly/dev builds... Hmm.

madrazo:
OK, @jcward @thomasuster  this is what I did 
Downloaded zip from https://github.com/thomasuster/hxcpp/tree/threaded-getCachedObjName
haxelib dev hxcpp hxcpp-threaded-getCachedObjName
tried to build a NME project, hxcpp asks to rebuild, choose yes and the following error occurs: 
_Uncaught exception - load.c(181) : Module not found : hxcpp-fallback.n_

git clone git@github.com:thomasuster/haxe-threadpool.git
gave the error: 
_Cloning into 'haxe-threadpool'...
Warning: Permanently added the RSA host key for IP address '192.30.253.112' to the list of known hosts.
Permission denied (publickey).
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists._

Downloaded zip from https://github.com/thomasuster/haxe-threadpool and 
haxelib dev haxe-threadpool haxe-threadpool-master

Tried building nme project, but again:
_Uncaught exception - load.c(181) : Module not found : hxcpp-fallback.n_

Removed hxcpp-threaded-getCachedObjName and copy again
Now asked again to rebuild the tools.

Compiled OK

thomasuster:
Thanks for testing @madrazo ! Were you able to reproduce the malloc error?

@jcward Hey! I modified the above a little bit to use https for clone instead of git, as well as the build instructions. This PR just modified the HXCPP build tool, so rebuilding is just...

```
cd hxcpp/tools/hxcpp
haxe compile.hxml
```

madrazo:
@thomasuster so far no issue, only this warning:
_hxcpp-threaded-getcachedobjname\include\Array.h(351): warning C4081: expected 'identifier'; found 'static'_
I'm using the official Haxe 3.4.2 , nme git

thomasuster:
@hughsando Hey Hugh, thoughts?

hughsando:
I like both ideas - multi-threaded hashes and cpp tool.
There are are a few things I think I will need to change.
- there is already a list of threads used for compiling - these should be combined with the thread pool if a thread pool is required (rather than just restarting new threads).
- since the thread pool code can be quite small, I would prefer a copy+paste over a dependency
- the thread pool should be message based, rather than mutex based
- the compiler should not be cloned, but maybe a thread-local version of the hashes kept, and a global set updated via a mutex.

I do not want to ship an additional cpp exe, so I think the logic might be that if the cpp exe exists, then it is used, and if it is does not, then it carries on with the neko version.  The developer would then optionally compile the tool with cpp is they need extra speed.

thomasuster:
1. combine threads ✅ 
2. copy paste over dependency ✅ 
3. message based: For performance? What's the reason for this?
4. compiler should not be cloned ✅ 

The fallback logic for non CPP tool should work with current logic. What should I adjust so that extra bin isn't shipped?

thomasuster:
Looks like OSX passed, but linux & windows are failing. Any ideas before I dive in? @hughsando ?

thomasuster:
@hughsando I tracked down the malloc error problem. This commit fixes it. 

Is this free intended to be here?

https://github.com/HaxeFoundation/hxcpp/pull/620/commits/9408c56e4f65c87536ef93b514bda4143c1436b3

thomasuster:
This PR depends upon https://github.com/HaxeFoundation/hxcpp/pull/631

Please hold off on merging this until that PR is resolved, as I might need to rebase this branch again to avoid duplicate commits.

thomasuster:
Here's the latest benchmarks...

Building pakka with all HXCPP_COMPILE_CACHE files in cache from master branch versus this branch.

HaxeFoundation:master
old-cached

thomasuster:threaded-getCachedObjName
new-cached

thomasuster:
Found bug in dependent PR, closing for now...

PR #627, opened 2017-08-29
xlocale.h was removed in glibc 2.26

ibilon:
My linux distribution upgraded to it (openSUSE tumbleweed) and now I can't build if I use `Sys`.

See 4.1 in https://sourceware.org/glibc/wiki/Release/2.26#Removal_of_.27xlocale.h.27

They mention that `xlocale.h` is a subset of `locale.h` so it should be fine, most tests passes locally (what doesn't doesn't seem related) for me but not sure what's the effect on earlier glibc.

Side note:
I get a lot of warning about `warning: dynamic exception specifications are deprecated in C++11 [-Wdeprecated]` `note: in expansion of macro ‘THROWS’`.
And the error was `Error: ./src/Ops.cpp: In static member function ‘static void Ops_obj::main()’:
./src/Ops.cpp:1601:85: warning: ISO C++ says that these are ambiguous, even though the worst conversion for the first is better than the worst conversion for the second:
 HXLINE(3355)  uint8 = (uint8 + anon2->__Field(HX_("xyz",59,78,5b,00),hx::paccDynamic));`
I have gcc/g++ 7.1.1

### DIFF
--- a/project/libs/std/Sys.cpp
+++ b/project/libs/std/Sys.cpp
@@ -30,7 +30,11 @@ int __sys_prims() { return 0; }
 #ifndef ANDROID
 #	include <locale.h>
 #if !defined(BLACKBERRY) && !defined(EPPC) && !defined(GCW0)
-#	include <locale.h>
+#	if defined(HX_LINUX)
+#		include <locale.h>
+#	else
+#		include <xlocale.h>
+#	endif
 #endif
 #endif
 #endif
--- a/project/libs/std/Sys.cpp
+++ b/project/libs/std/Sys.cpp
@@ -30,7 +30,7 @@ int __sys_prims() { return 0; }
 #ifndef ANDROID
 #	include <locale.h>
 #if !defined(BLACKBERRY) && !defined(EPPC) && !defined(GCW0)
-#	include <xlocale.h>
+#	include <locale.h>
 #endif
 #endif
 #endif
--- a/src/hx/libs/std/Sys.cpp
+++ b/src/hx/libs/std/Sys.cpp
@@ -31,7 +31,7 @@
    #ifndef ANDROID
       #include <locale.h>
       #if !defined(BLACKBERRY) && !defined(EPPC) && !defined(GCW0)
-         #include <xlocale.h>
+         #include <locale.h>
       #endif
    #endif
 #endif
--- a/src/hx/libs/std/Sys.cpp
+++ b/src/hx/libs/std/Sys.cpp
@@ -31,7 +31,11 @@
    #ifndef ANDROID
       #include <locale.h>
       #if !defined(BLACKBERRY) && !defined(EPPC) && !defined(GCW0)
-         #include <locale.h>
+        #if defined(HX_LINUX)
+            #include <locale.h>
+        #else
+            #include <xlocale.h>
+        #endif
       #endif
    #endif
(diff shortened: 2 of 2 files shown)


andyli:
It doesn't work on Mac, as shown in Travis:
```
Error: /Users/travis/build/HaxeFoundation/hxcpp/src/hx/libs/std/Sys.cpp:146:4: error: unknown type name 'locale_t'
   locale_t lc, old;
   ^
/Users/travis/build/HaxeFoundation/hxcpp/src/hx/libs/std/Sys.cpp:147:19: error: use of undeclared identifier 'LC_TIME_MASK'
   lc = newlocale(LC_TIME_MASK,l.__s,NULL);
                  ^
/Users/travis/build/HaxeFoundation/hxcpp/src/hx/libs/std/Sys.cpp:150:4: error: use of undeclared identifier 'old'
   old = uselocale(lc);
   ^
/Users/travis/build/HaxeFoundation/hxcpp/src/hx/libs/std/Sys.cpp:151:9: error: use of undeclared identifier 'old'
   if( !old )
        ^
/Users/travis/build/HaxeFoundation/hxcpp/src/hx/libs/std/Sys.cpp:156:8: error: use of undeclared identifier 'old'
   if( old != LC_GLOBAL_LOCALE )
       ^
/Users/travis/build/HaxeFoundation/hxcpp/src/hx/libs/std/Sys.cpp:156:15: error: use of undeclared identifier 'LC_GLOBAL_LOCALE'
   if( old != LC_GLOBAL_LOCALE )
              ^
/Users/travis/build/HaxeFoundation/hxcpp/src/hx/libs/std/Sys.cpp:157:18: error: use of undeclared identifier 'old'
      freelocale(old);
                 ^
7 errors generated.
```

The way I solved it for neko is to detect the present of xlocale.h during configuration:
https://github.com/HaxeFoundation/neko/commit/e286c8f3301fa443255a3101d14b7392467051b8
I'm not sure if there is a way to do that in hxcpp?

ibilon:
Seems like mac doesn't use glibc, maybe xlocale should be included for it and locale for linux?

andyli:
I'm not sure. The Mac behavior could be caused by the difference between clang and gcc. That means we should not detect OS, but the compiler in use. Have to try use clang on Linux and use gcc on Mac and see the result...

hughsando:
I think if we use HX_LINUX to use local instead of xlocal we should be ok.
Theoretically, mac + new gcc could have the problem too, but without a test of this config I would not like to speculate. A simple linux test should leave it no worse - unless some of the linux-like environments (android) break.  If this happens i think a new approach might be to just paste the constants into the cpp file and skip the header altogether, which is how I have to deal with some android issues.

andyli:
Just found a nice trick in SO: https://stackoverflow.com/a/142995/267998
Basically write our own xlocale.h and make it to be in the last item of search paths. So the cpp compiler will use the system xlocale.h if it's available or fallback to our custom xlocale.h.

ibilon:
Couldn't find how to use hxcpp+clang on linux and don't have a mac,
so made the change to use `HX_LINUX`.

> Just found a nice trick in SO: https://stackoverflow.com/a/142995/267998

Would work, but technically you're not supposed to use xlocale with glibc, it's not part of the public api, that's why they removed it.

PR #628, opened 2017-08-29
Android NDK detection adjust

madrazo:
-Use getNdkVersion to check version in both source.properties file and
name of directory.  It returns a float value that can be compared.
-Try to use the default ndk-bundle from SDK manager if there is no valid
NDK in ANDROID_NDK_DIR
-fix NDKVXX flag to force looking for NDKs even if ANDROID_NDK_ROOT is
set

### DIFF
--- a/tools/hxcpp/Setup.hx
+++ b/tools/hxcpp/Setup.hx
@@ -5,40 +5,59 @@ import BuildTool;
 
 class Setup
 {
-   static function findAndroidNdkRoot(inDir:String, inBaseVersion:Int):String
+   static function findAndroidNdkRoot(defines: Map<String,String>, inBaseVersion:Int):String
    {
-      var files:Array<String> = null;
-      try
-      {
-         files = FileSystem.readDirectory(inDir);
-      }
-      catch (e:Dynamic)
-      {
-         Log.warn('ANDROID_NDK_DIR "$inDir" does not point to a valid directory');
-         return null;
-      }
-
       var bestVersion = 0.0;
       var result:String = null;
-      for (file in files)
-      {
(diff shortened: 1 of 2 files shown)


tools/hxcpp/Setup.hx
```diff
@@ -5,82 +5,83 @@ import BuildTool;
...
+        try
+        {
+             files = FileSystem.readDirectory(ndkDir);
+        }
+        catch (e:Dynamic)
+        {
+             Log.warn('ANDROID_NDK_DIR "$ndkDir" does not point to a valid directory');
+             checkFiles=false;
+        }
+        if (checkFiles)
```
thomasuster: Is this bool needed anymore? Seems like you could just call a function to check the files inside of the catch.

madrazo: Initialy this code only considered the name of directories to detect the NDK version, and later added to check source.properties file.  The check is done in 2 places, but now it calls getNdkVersion on both. Things start to be a little tricky when trying to use the NDKVXX flag. I believe this code fixes it when retrieving the version from the source.properties file and also gives a higher prority than ANDROID_NDK_ROOT. It also checks in the default Android studio ndk-bundle directory.

madrazo: Yes, maybe it would be the same as writting the checkFiles block inside the "try"

Gama11:
These changes cause Android compilation to break for me, see #715.

madrazo:
Something that is different:
var extract_version = ~/^(android-ndk-)?r(\d+)([a-z]?)$/;

var extract_version = ~/.*\/(android-ndk-)?r(\d+)/;    

Please give it a try

PR #629, opened 2017-09-03
Do not include xlocale.h on systems, where __GLIBC__ is defined.

dmacvicar:
Fixes build on glibc 2.26, where xlocale was removed.
https://sourceware.org/git/?p=glibc.git;a=commit;h=f0be25b

Should not be needed in glibc systems:
https://lists.opensuse.org/opensuse-factory/2017-08/msg00580.html
>xlocale.h has been obsolete for nearly 10 years, since POSIX.1-2008.


ibilon:
Seems like a better fix than mine https://github.com/HaxeFoundation/hxcpp/pull/627 :D

PR #631, opened 2017-09-07
Added stress test that fails

thomasuster:
Hey Hugh

I narrowed down another test case that fails for me. It also seemed to give me the same malloc error that I've been unable to track down when trying to multithread the build tool & building large test projects.

Any ideas?


hughsando:
There could be a problem with the number of open file handles, or some other resource (number of threads/memory  - since each thread is going to need a stack)
10000 threads seems like a crazy number to have live.  I wonder if a look of 100 runs of 100 threads would have the same problem?

thomasuster:
I'll modify this one in that way later tonight, good idea.

thomasuster:
@hughsando Updated. This repro's the malloc free error instead of seg faults, so this is a probably much better test.

screen shot 2017-09-07 at 7 52 07 pm

thomasuster:
The bug returns... Closing for now.

PR #635, opened 2017-09-17
Flakiness & Stress tests

thomasuster:
Tough GC bugs seem to be more probabilistic than deterministic, to help with this I added the ability to allow haxe tests to be run 10 times to test if things are flaky. This passed for existing upstream/master.
```
cd test/haxe
haxe compile.hxml -D FLAKINESS_TEST; bin/TestMain
```

I then added a new unit test to reproduce the bug I was seeing in my other PR that I'm not putting on hold until this all seems solid.

```
cd test/haxe
haxe compile.hxml -D STRESS_TEST; bin/TestMain
```
This would fail 99% of the time. I tracked this down to what I believe is a race condition with threads & large block recycles. I believe without locking a mutex before iterating largeObjectRecycle then it's possible for multiple threads to grab the same large object, so then it gets added back twice to the list, and then doubled freed on the next Collect.

My fix with the lock now passes
```
haxe compile.hxml -D STRESS_TEST; bin/TestMain
```
but now was failing 
```
haxe compile.hxml -D FLAKINESS_TEST; bin/TestMain
```

I determined that it wasn't the fix causing the problem, but merely running the new test...
hxcpp/test/haxe/gc/TestGCWithSys.hx

Even if you adjust
```haxe
doThreadedWork(4,1000);
```
to
```haxe
doThreadedWork(4,4);
```
It somehow produces a side effect that causes the subsequent tests to fail during a FLAKINESS_TEST.

I was able to fix this by disabling HXCPP_GC_MOVING inside...
hxcpp/test/haxe/compile.hxml

So now that FLAKINESS_TEST and STRESS_TEST individually work I naturally tried...
```
cd test/haxe
haxe compile.hxml -D STRESS_TEST -D FLAKINESS_TEST; bin/TestMain
```
This fails! Very unfortunately this feels like a different bug related to closures. It looks like variables closures are referencing may be getting cleaned up before closures are done using them.

Because these new flags STRESS_TEST/FLAKINESS_TEST are internal, I can very easily load this into lldb to get reasonable back traces.
```
* thread #2, stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
  * frame #0: 0x000000010008c57b TestMain`String::String(Dynamic const&) + 27
    frame #1: 0x000000010002141d TestMain`gc::TestGCWithSys_obj::makeThreads(int)::_hx_Closure_0::_hx_run() + 173
    frame #2: 0x0000000100021341 TestMain`gc::TestGCWithSys_obj::makeThreads(int)::_hx_Closure_0::__run() + 17
    frame #3: 0x0000000100067cf7 TestMain`hxThreadFunc(void*) + 151
    frame #4: 0x00007fffb788193b libsystem_pthread.dylib`_pthread_body + 180
    frame #5: 0x00007fffb7881887 libsystem_pthread.dylib`_pthread_start + 286
    frame #6: 0x00007fffb788108d libsystem_pthread.dylib`thread_start + 13
```


hughsando:
I have fixed 2 bugs - one in the locking, and one when old threads get brought back to life without clearing.  I have also merged your test in - you can use a command-line arg to run more times.  I changed a few things around, and did a manual merge.  Thanks.
https://github.com/HaxeFoundation/hxcpp/commit/654a77257148115218d2e4d09d0a4c86644d30aa

PR #640, opened 2017-09-22
Threaded get cached obj name

thomasuster:
* Increases performances of caches builds by switching build tool to CPP and multi-threading key calculations
* Haxe trunk removed haxe/unit, so I added it to each test src for now

Benchmark Check
This is the results of a build after already building the project.

Current HXCPP: 24.2s
old-cached

Multithreaded CPP HXCPP: 16.2s
new-cached

33% faster! Not bad.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ test/cffi/project/ndll
 test/snippets/messagebox/cpp
 test/haxe/gc/big.txt
 hxcpp.n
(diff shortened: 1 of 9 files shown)


Gama11:
Should probably use [hx3compat](https://github.com/HaxeFoundation/hx3compat) instead of copying `haxe.unit`? :)

thomasuster:
@hughsando bump

Issue #621, opened 2017-08-04
Error: conversion from 'long' to 'const Dynamic' is ambiguous


thejustinwalsh:
When [compiling SiON](https://github.com/thejustinwalsh/SiON) for macOS (`lime build mac -debug`)  I am getting the following compiler error:
```cpp
Error: ./src/org/si/sion/sequencer/base/MMLParser.cpp:756:45: error: conversion from 'long' to 'const Dynamic' is ambiguous
int defaultValue = __o_defaultValue.Default(-2147483648);
                                            ^~~~~~~~~~~
/usr/local/lib/haxe/lib/hxcpp/3,4,64/include/Dynamic.h:21:4: note: candidate constructor
   Dynamic(int inVal);
   ^
/usr/local/lib/haxe/lib/hxcpp/3,4,64/include/Dynamic.h:22:4: note: candidate constructor
   Dynamic(short inVal);
   ^
/usr/local/lib/haxe/lib/hxcpp/3,4,64/include/Dynamic.h:23:4: note: candidate constructor
   Dynamic(unsigned int inVal);
   ^
/usr/local/lib/haxe/lib/hxcpp/3,4,64/include/Dynamic.h:24:4: note: candidate constructor
   Dynamic(unsigned short inVal);
   ^
/usr/local/lib/haxe/lib/hxcpp/3,4,64/include/Dynamic.h:25:4: note: candidate constructor
   Dynamic(unsigned char inVal);
   ^
/usr/local/lib/haxe/lib/hxcpp/3,4,64/include/Dynamic.h:26:4: note: candidate constructor
   Dynamic(signed char inVal);
   ^
/usr/local/lib/haxe/lib/hxcpp/3,4,64/include/Dynamic.h:28:4: note: candidate constructor
   Dynamic(bool inVal);
   ^
/usr/local/lib/haxe/lib/hxcpp/3,4,64/include/Dynamic.h:29:4: note: candidate constructor
   Dynamic(double inVal);
   ^
/usr/local/lib/haxe/lib/hxcpp/3,4,64/include/Dynamic.h:30:4: note: candidate constructor
   Dynamic(float inVal);
   ^
/usr/local/lib/haxe/lib/hxcpp/3,4,64/include/Dynamic.h:31:4: note: candidate constructor
   Dynamic(cpp::Int64 inVal);
   ^
/usr/local/lib/haxe/lib/hxcpp/3,4,64/include/Dynamic.h:32:4: note: candidate constructor
[...]

PR #667, opened 2017-10-31
Dynamic constructors for long and unsigned long

thejustinwalsh:
Resolves an `error: conversion from 'long' to 'const Dynamic' is ambiguous` when using singed integer literals in a haxe program.

### DIFF
--- a/include/Dynamic.h
+++ b/include/Dynamic.h
@@ -18,8 +18,10 @@ class HXCPP_EXTERN_CLASS_ATTRIBUTES Dynamic : public hx::ObjectPtr<hx::Object>
 public:
 
    Dynamic() {};
+   Dynamic(long inVal);
    Dynamic(int inVal);
    Dynamic(short inVal);
+   Dynamic(unsigned long inVal);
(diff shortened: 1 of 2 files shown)


thejustinwalsh:
Resolves issue #621

thejustinwalsh:
Well, Xcode failed complaining about Homebrew and the version of ruby that is installed...
And linux thinks that `cpp::Int64` and `long` are trying to overload each other, curiously this isn't consistent on all platforms.

How do you feel about just moving to [Fixed width integer types](http://en.cppreference.com/w/cpp/types/integer)? Is C++11 optional or required at this point?

aW4KeNiNG:
I have this problem in Mac too: `error: conversion from 'size_t' (aka 'unsigned long') to '::Dynamic' is ambiguous`

```
var num:cpp.SizeT = 10;
trace('My num is ${num}');
```

The PR was changed in https://github.com/HaxeFoundation/hxcpp/commit/d9c02ec114f7db00070b46a91a45ff3bfc353530

PD: In Windows works fine.

PR #694, opened 2018-02-15
the unicode changes aren't complete without these changes too

larsiusprime:
I was getting some errors without them. Also FileSystem.exists() still doesn't seem to work on a file name with unicode letters in it, even though both sys_exists implementations seem to be using wide character strings


larsiusprime:
Whoops, looks like I got the branches tangled up and these are already fixed on the main one. Nevermind!

PR #697, opened 2018-02-23
Add Array.resize method

bendmorris:
From https://github.com/HaxeFoundation/haxe/issues/6173

Adds a `resize` method to Array. Didn't implement this for cppia as I'm not familiar with that target.

### DIFF
--- a/include/Array.h
+++ b/include/Array.h
@@ -233,6 +233,7 @@ public:
    inline Dynamic __zero(const Dynamic &a0,const Dynamic &a1)  { zero(a0,a1); return null(); }
    virtual Dynamic __memcmp(const Dynamic &a0) = 0;
    virtual void __qsort(Dynamic inCompare) = 0;
+   virtual Dynamic __resize(const Dynamic &a0) = 0;
 
    #else
    inline void ____SetSize(int len)  { __SetSize(len); } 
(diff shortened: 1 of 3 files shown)


hughsando:
This is currently done via:
```haxe
using cpp.NativeArray;

...
array.setSize(1234);
```
How were you intending to call 'resize' - dynamically or untyped?

bendmorris:
See the linked Haxe PR - intention is to add this as a standard Array method.

bendmorris:
@hughsando - any issues with this implementation? Is there a better way to do it?

bendmorris:
@hughsando - bump. The goal here is to add Array.resize for all targets, and unless that's possible without modifying hxcpp, this PR would need to go in first. Would appreciate a response.

hughsando:
I guess my main issue is to do with the removal or wrapping of the `__SetSize` function.
If `resize` is going to be official, I think all the setSize calls should be replaced with resize calls,  and there should be some setSize stubs to call resize for backwards compatibility.
Also, I think the VirtualArray version is going to need empty array condition that is in the `__SetSize` function - although this is not an issue if the set size function is renamed.
Similarly for cppia - will need a solution here, or the tests will fail - it will be very similar to the SetSizeExact that is already in there.
Other questions are - the return value - void or 'this' ?
And, is it `__SetSizeExact` or `__SetSize` - the difference being the exact version will free memory if you shrink the array, whereas SetSize will not.
None of these issues is that big or difficult - they are pretty trivial really. But it was enough to stop me just hitting the merge button, and then I kinda forgot about the issue.

hughsando:
Ok, I see from the other targets that no memory is released, and it returns void, so it is for `__SetSize`

bendmorris:
Thanks Hugh. I'll look into cppia next.

hughsando:
I put the changes into cppia, but have not really tested them.

bendmorris:
Awesome - my haxe PR includes tests, so I'll get back to you if there are any issues.

PR #707, opened 2018-05-10
Progress on HXCPP_CLANG on Android (Don't merge)

jgranick:
Android NDK v16 and 17 no longer support GCC. I have been working on trying to add clang support to the Android toolchain, but have run into some trouble. I'm posting my progress here, perhaps someone else will find what I am missing?


jgranick:
I currently have this error:

```
In file included from lib/sdl/include\SDL.h:32:
In file included from lib/sdl/include/SDL_main.h:25:
In file included from lib/sdl/include/SDL_stdinc.h:84:
In file included from D:\Development\Android\android-ndk-r17/sources/android/support/include\math.h:32:
D:\Development\Android\android-ndk-r17/sources/cxx-stl/llvm-libc++/include\math.h:1302:93: error: no member named 'log2f' in the global namespace
inline _LIBCPP_INLINE_VISIBILITY float       log2(float __lcpp_x) _NOEXCEPT       {return ::log2f(__lcpp_x);}
```

This probably has to do with including the standard lib properly

nanjizal:
Do you have lastest NDK I would guess this might be related?
https://github.com/MIPS/ndk/commit/a74e525c629d84e634fd7242053172333a3711d5

jgranick:
I don't think so, if it were missing in the shared library, we would have an error when linking. The error is while compiling, so something about the headers is not correct, I assume

hughsando:
Looks like a path include-order thing. If you search the headers for log2f,
you might find a directory that needs to be up the include path.
If you can compile from the standard android tools, you can put a "#error"
in the files an this gives reasonable debug about what the standard tools
do.

hughsando:
Anthon's link looks like it might be googles fault.  You might be able to
fix  it  by adding the android  prototype to the global namespace somewhere
that hxcpp includes, like hx/StdLibs.h,  then implement it in
AndroidCompat.cpp - or just implement it inline in the header.
If the SDL includes hxcpp.h first, this should fix it.
Not sure if there are more than just this function.

Dimous:
Have you made any progress on this?

jgranick:
I just tried again and made additional progress:

Rebuilding the Lime shared library I get only one error:

```
/home/joshua/Development/Android/android-ndk-r19c/platforms/android-16/arch-arm/usr/lib/../lib/crtbegin_dynamic.o:crtbegin.c:function _start_main: error: undefined reference to 'main'
```

I am not sure how to force it to use a different CRT object or how to not look for main?

Gama11:
With https://github.com/HaxeFoundation/hxcpp/pull/825 merged, this can probably be closed?

hughsando:
I'm closing this now, since I think clang is working because of other changes.

PR #708, opened 2018-05-12
Fix #701 - haxe.Log.trace(msg, null)

### LINKED ISSUE #701: Trace with null position traces "?:0"
[From the docs](https://haxe.org/manual/debugging-trace-log.html): "To trace without the default position information haxe.Log.trace(msg, null) can be used."

On C++ this results in:

`?:0: blah blah blah`

robocoder:
Proposed patch for review/feedback.

Not sure how to write a unit test for this.

### DIFF
--- a/src/hx/StdLibs.cpp
+++ b/src/hx/StdLibs.cpp
@@ -20,7 +20,7 @@ typedef int64_t __int64;
 #include <syslog.h>
 #endif
 #ifdef TIZEN
-extern "C" EXPORT_EXTRA void AppLogInternal(const char* pFunction, int lineNumber, const char* pFormat, ...);
+#include <dlog.h>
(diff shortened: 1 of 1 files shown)


robocoder:
In the when all `#else` fails case, I changed `%S` to `%ls` for portability but upon further examination, perhaps this should simply be `%s` as I don't see how `x.to string().__s` returns a wide char string.

Gama11:
You must've accidentally changed the line endings of `StdLibs.cpp`, it shows 804 additions and 792 deletions. If you add `?w=1` to the diff it becomes readable: https://github.com/HaxeFoundation/hxcpp/pull/708/files?w=1

hughsando:
I hop this is fixed now.

PR #724, opened 2018-08-25
Directory structure of "libstdc++" is different

thefurfel:
```
user@mypc MINGW64 /d/Android/ndk-bundle/sources/cxx-stl/gnu-libstdc++/4.9/libs
$ ls
arm64-v8a/  armeabi-v7a/  mips/  mips64/  x86/  x86_64/
```


Furfel:
(As of latest NDK v17 on Windows)

madrazo:
"armabi" and "armabi-v7" are different kind of ABIs. It seems that support for armabi was dropped in NDK 16.

Simn:
Given that this PR no longer merges, I'll go ahead and close it.

PR #736, opened 2018-10-29
Add explicit haxe path

stevemayhew:
This change basically works the same as before (cloning the latest change from HaxeFoundation/haxe).  I would consider going one step further and specifying an explict tag/commit from haxe.

With this your build tree will have stable test results even if the haxe repository changes.


hughsando:
So, to check, this should not really effect me, but it allows you to use a fixed haxe version?

Simn:
I could see this being useful, but the branch isn't mergeable anymore and there was no follow-up, so I'm closing.

Issue #706, opened 2018-05-09
Please support latest ndk


ghost:
See https://github.com/openfl/lime/issues/1179

hughsando:
I think this has been fixed?
I am closing this now, but feel free to re-open it.

Gama11:
By what? https://github.com/HaxeFoundation/hxcpp/pull/707 is still open.

hughsando:
I thought by this:  https://github.com/openfl/lime/issues/1179#issuecomment-387901554
 And my understanding (according to the title) was that progress on https://github.com/HaxeFoundation/hxcpp/pull/707 was stopped.

If it's google's problem again, we could wait for them.  Or I could add something to my "google don't give a crap about compatibility hack" file.  I have not tried the latest NDK yet.

Gama11:
There's two different issues discussed in https://github.com/openfl/lime/issues/1179, one was a Lime issue and fixed since. The other one is that the Android NDK seems to be dropping GCC support starting with version 16 (as mentioned in #707 by @jgranick), so HXCPP will need to be able to build with clang instead. The current workaround is to instruct people to downgrade to NDK r15c, which isn't ideal. :)

Gama11:
Well, perhaps they haven't _dropped_ GCC support just yet? Seems surprisingly hard to find an official statement on it, but it's definitely deprecated and no longer updated, so sooner or later the switch to clang has to happen. Relevant: https://android-developers.googleblog.com/2017/09/introducing-android-native-development.html

Gama11:
Ah, the roadmap has something more concrete:

>NDK r18
>Estimated release: Q2 2018
>
> [...]
>
>Remove GCC
>GCC is still in the NDK today because some of gnustl's C++11 features were written such that they do not work with Clang (threading and atomics, mostly). Now that libc++ is the best choice of STL, this is no longer blocking, so GCC can be removed.

https://android.googlesource.com/platform/ndk/+/master/docs/Roadmap.md#ndk-r18

Gama11:
Could we maybe implement a helpful error message like _"r15c or older required"_ when trying to use a version that's too recent? I've lost count of how many people I've had to tell to downgrade to r15c. :)

joshtynjala:
[...]

PR #742, opened 2018-11-15
Try to enable NDK 18 compile (Don't merge)

madrazo:
It compiles but still has linking/running errors. Use it as reference.
Check https://developer.android.com/ndk/guides/standalone_toolchain

tested with NME DisplayinABitmap: `nme android -gradle -DPLATFORM=android-26`

https://github.com/HaxeFoundation/hxcpp/issues/706
https://github.com/HaxeFoundation/hxcpp/pull/707


Simn:
This one says "Don't merge" and has conflicts, so I assume that it's safe to close at this point.

PR #744, opened 2018-11-19
Add HXCPP_GC_SINGLE_THREADED define

Sunjammer:
This simply adds a define to constrain MAX_MARK_THREADS to 1.

Note, this PR is in motion. I welcome any guidance on best approaching this problem if my impl is not enough or a bad call.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,4 +1,10 @@
 
+
+* Throw exceptions according to the spec when casting
+* Introduce hxcpp_smart_strings for unicode text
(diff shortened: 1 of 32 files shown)


hughsando:
Hi.  It is not clear what the actual changes you are looking to apply are  Do you think you could rebase this agast the current version, or do a new PR with just the files(s) that you are interested in?

Sunjammer:
Outdated

PR #752, opened 2018-12-05
propagate nullness of Array<T> to Array<Dynamic>

### LINKED ISSUE #746: hxcpp fails to retain null-ness of cast Array
When a function is called and an Array instance is implictly cast, if the Array instance was null in the caller's frame, it is not null in the callee's frame, instead becoming an empty array.  Here is an example showing the problem:

```
class HxcppBug
{
    public static function main()
    {
        var a : Array<Int> = null;

        examine(a);
    }

    private static function examine(a : Array<Dynamic>)
    {
        if (a != null) {
            trace("TEST FAILED: a was null when passed in but is not null " +
                  "inside examine");
        }
    }
}
```

This bug is pretty egregious.

ncannasse:
fixed Enum.getParameters() should not return null
close #746

### DIFF
--- a/include/cpp/VirtualArray.h
+++ b/include/cpp/VirtualArray.h
@@ -562,7 +562,7 @@ public:
 // Build dynamic array from foreign array
 template<typename SOURCE_>
 VirtualArray::VirtualArray( const Array<SOURCE_> &inRHS )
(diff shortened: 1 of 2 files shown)


bjitivo:
Thank you, looking at my local code base I see that I had already applied exactly this fix to our local version.

PR #758, opened 2019-01-15
Avoid entering GCFreeZone twice

RobDangerous:
For some reason this caused trouble for me on Switch but nowhere else but seems appropritate to fix that in any case.


RobDangerous:
Actual problem turned out to be something else entirely which just randomly crashed at that position. I still like this pull request though.

PR #768, opened 2019-03-01
mbedtls-2.16.0 & Safe Calls

codeservice:


Aurel300:
I think you might want to do two PRs, one for the calls, one for TLS library (?). For TLS, please mark external code as vendored and ideally in a separate commit (or multiple). This way it is not very friendly to code review.

codeservice:
Yes, to many changes, not clean. Closing it for breaking on two.

PR #775, opened 2019-03-31
Correct static lib paths for hxcpp 4

haxiomic:
Seems like hxcpp static libs have moved from bin/ to lib/, this PR updates the example with the location change

I've tested on mac and the instructions work, but not tested on Windows so I'm assuming the windows lib paths are correct in this PR


haxiomic:
This is non-critical so I'll close to reduce PR clutter

PR #779, opened 2019-04-06
[android] Fix NDK detection regex

### LINKED ISSUE #778: [android] Cannot deduce NDK
```
Warning: Could not deduce NDK version from "android-ndk-r19c" - assuming 8
Warning: Could not deduce NDK version from "/Users/thomasuster/SDKs//android-sdk/ndk-bundle" - assuming 8
```

thomasuster:
Fixes #778

### DIFF
--- a/tools/hxcpp/Setup.hx
+++ b/tools/hxcpp/Setup.hx
@@ -96,7 +96,7 @@ class Setup
 
       var dir = inDirName.split("\\").join("/");
       Log.v('Try to get version from directory name "$dir"');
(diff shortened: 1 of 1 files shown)


thomasuster:
https://try.haxe.org/#F78f2

hughsando:
I guess we could get rid of the slash condition altogether for the same effect - and the android-ndk- bit too if it is optional. (eg, "far19c" now matches).
Maybe the input should be a full path?

PR #781, opened 2019-04-15
Unicode fixes

Aurel300:
This fixes some of the issues with Unicode support, particularly on non-Windows platforms. "Smart strings" currently represent a string as either a `char *` when there are no non-ASCII characters, or otherwise a `char16_t *` (UTF-16 codepoints). Unfortunately, it seems there is still no stdlib function to print a `char16_t *` in C++, only `wchar_t *` using `%S`. Additionally, `wchar_t` does not have a specified size (it is 16 bits on Windows, 32 bits on Linux and OS X).

To fix platforms where `sizeof(char16_t) != sizeof(wchar_t)`, we could:

 - change the internal representation for Unicode strings to use `wchar_t *` - this would use the "proper" type for the strings but could introduce inconsistencies when dealing with non-BMP characters; also strings would take twice as much memory on Linux and OS X
 - convert `char16_t *` to `wchar_t *` in a temporary buffer just before outputting a Unicode string

This PR chooses the latter approach.

## TODO

 - [x] `setlocale` re-enabled - this was commented out in a `#if defined(_MSC_VER) && !defined(HX_WINRT)` block
   - [ ] check compatibility with Windows
 - [x] `char16_t *` to `wchar_t *` conversion function - `__hxcpp_utf16_to_wchar` in `StdLibs.cpp`
 - [x] fix `trace`
 - [x] fix `Sys.print`, `Sys.println`
 - [x] get rid of any remaining occurrences of `HX_UTF8_STRINGS` (should now be obsolete)
 - [x] smart strings are ignored with `HX_WINRT`, `defined(TIZEN)`, `defined(HX_ANDROID) && !defined(HXCPP_EXE_LINK)`, or `defined(WEBOS)` - the defines mostly change which function is used in place of `printf`; if they all support `%S`, the `ifdef`s should only define the `printf` function (already done for `snprintf` in `String.cpp`)
 - [x] ~~`Bytes.writeString`~~ `Output.writeString` just uses `Bytes.ofString`, which encodes the string correctly (UTF-16 into UTF-8) via `__hxcpp_bytes_of_string`
 - [x] fix `String.split` on Unicode string with non-Unicode delimiter
 - [ ] possible issues with `Bytes.getString`
   - `__hxcpp_string_of_bytes` falls back to `String::fromCharCode` for any 1-character string - see below for inconsistent results
   - fix `_hx_utf8_decode_advance` / `DecodeAdvanceUTF8` ?
     - currently reads OOB (e.g. if the last byte is `F0`, three extra bytes are read) - add an `end` argument? - would break any code using `cpp.NativeString.utf8DecodeAdvance` (assuming there even is any)
     - no error handling for invalid sequences - lenient decoder by default, use `UTF8Strict` to catch errors

## `__hxcpp_string_of_bytes` ##

```haxe
class Main {
  public static function main():Void {
    var s = [
      haxe.io.Bytes.ofHex("F3").getString(0, 1),
      haxe.io.Bytes.ofHex("60F3").getString(0, 2),
      haxe.io.Bytes.ofHex("6060F3").getString(0, 3),
      haxe.io.Bytes.ofHex("606060F3").getString(0, 4),
      haxe.io.Bytes.ofHex("60606060F3").getString(0, 5),
      haxe.io.Bytes.ofHex("6060606060F3").getString(0, 6),
      haxe.io.Bytes.ofHex("606060606060F3").getString(0, 7),
      haxe.io.Bytes.ofHex("60606060606060F3").getString(0, 8),
      haxe.io.Bytes.ofHex("6060606060606060F3").getString(0, 9),
      haxe.io.Bytes.ofHex("606060606060606060F3").getString(0, 10),
      haxe.io.Bytes.ofHex("60606060606060606060F3").getString(0, 11)
    ];
    for (i in 0...s.length) trace(s[i].length, s[i].charCodeAt(s[i].length - 1));
  }
}
```

Probably reads OOB / uninitialised memory for all cases except the first. Ouputs are inconsistent:

```
Main.hx:16: 1,243
Main.hx:16: 3,56320
Main.hx:16: 4,56320
Main.hx:16: 5,56320
Main.hx:16: 6,56320
Main.hx:16: 7,56321
Main.hx:16: 8,56454
Main.hx:16: 9,56704
Main.hx:16: 10,56320
Main.hx:16: 11,56321
Main.hx:16: 12,56390
```

### DIFF
--- a/include/hxcpp.h
+++ b/include/hxcpp.h
@@ -48,8 +48,6 @@
 
 #include <string.h>
 
-#define HX_UTF8_STRINGS
-
 #include <wchar.h>
 
 #ifdef HX_LINUX
--- a/src/String.cpp
+++ b/src/String.cpp
@@ -1556,7 +1556,7 @@ Array<String> String::split(const String &inDelimiter) const
    Array<String> result(0,1);
    #if HX_SMART_STRINGS
    bool s0 = isUTF16Encoded();
-   bool s1 = isUTF16Encoded();
+   bool s1 = inDelimiter.isUTF16Encoded();
    if (s0 || s1)
    {
       if (s0 && s1)
--- a/src/String.cpp
+++ b/src/String.cpp
@@ -26,29 +26,17 @@ using namespace std;
 // vc 7...
 #if _MSC_VER < 1400 
 
-#ifdef HX_UTF8_STRINGS
 #define SPRINTF _snprintf
-#else
-#define SPRINTF _snwprintf
-#endif
 
 #else // vc8+
 
-#ifdef HX_UTF8_STRINGS
 #define SPRINTF _snprintf_s
-#else
-#define SPRINTF _snwprintf_s
-#endif
 
 #endif
 
 #else // not windows ..
 
-#ifdef HX_UTF8_STRINGS
 #define SPRINTF snprintf
-#else
-#define SPRINTF swmprintf
-#endif
 
 #endif
 
@@ -58,14 +46,9 @@ using namespace std;
 
 namespace hx
 {
-#ifdef HX_UTF8_STRINGS
 char HX_DOUBLE_PATTERN[20] = "%.15g";
 #define HX_INT_PATTERN "%d"
 #define HX_UINT_PATTERN "%ud"
-#else
-wchar_t HX_DOUBLE_PATTERN[20] =  L"%.15g";
-#define HX_INT_PATTERN L"%d"
-#endif
 }
 
 void __hxcpp_set_float_format(String inFormat)
@@ -1489,10 +1472,6 @@ const char16_t * String::wc_str() const
 
 const wchar_t * String::__WCStr() const
 {
-   #ifndef HX_UTF8_STRINGS
-   return __s ? __s : L"";
-   #else
-   
    const unsigned char *ptr = (const unsigned char *)__s;
    const unsigned char *end = ptr + length;
    int idx = 0;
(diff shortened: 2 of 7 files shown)


hughsando:
This is not a bad approach.  The other one I was thinking of was putwchar in a loop.  Not sure of performance,  avoiding the alloc might be nice, but it depends on how efficient the call is.

Aurel300:
> This is not a bad approach. The other one I was thinking of was putwchar in a loop. Not sure of performance, avoiding the alloc might be nice, but it depends on how efficient the call is.

I did think of that also, but I think performance-wise it would be much worse. Currently we have `setbuf(stdout, 0)` in place, so every call to `put(w)char` would be a syscall I believe. We could set the stdout buffer to be exactly the right size just before we do the print loop, but then we're just letting the stdlib do the exact same thing as the PR does atm (allocate a temporary buffer).

hughsando:
yes, sounds about right.
The Bytes.writeString it think should be char16_t to match javascript and others.

hughsando:
I have merged this since I am going to do some work in this area today.

hughsando:
I have made quite a few changes in this area - although I'm not 100% sure - I have decided to output utf8 to the stream, mainly because wchar_t output does not really seem to "just work" on windows so is of limited utility.  With utf8, you can redirect to a file, and viewing the file looks ok.  The localization of the console is still a bit of a mystery to me.
I have also introduced some buffering object to make the conversions easy - I might tweak these a bit to allow them to work on cffi too.

ncannasse:
@Aurel300 can you comment whereas the changes are passing the unit tests now, and if not what's still missing?

PR #782, opened 2019-04-22
Fix faulty detection of UTF-16 surrogates

Aurel300:
This should not throw:

```haxe
var a = haxe.io.Bytes.ofString(String.fromCharCode(0xE000));
```


hughsando:
I have allowed the high character codes > 0xe000 and implemented surrogate checking - replacing invalid values with 0xfffd, which I think should solve this problem.

PR #783, opened 2019-04-23
Sys.args with unicode

Aurel300:
More Unicode fixes…

 - [x] `dup()` fixed for UTF-16 strings
 - [ ] should `dupConst()` be fixed also? I'm not sure what the `const` refers to there
 - [x] fix `__get_args` if `_hxcpp_argc/v` is set
 - [x] fix `__get_args` on OS X
 - [ ] fix `__get_args` for Linux and Windows - these all add to a string character by character, even though the "characters" are actually partial Unicode bytes; some temporary buffer should be allocated and then decoded into a String I assume?

### DIFF
--- a/include/hxString.h
+++ b/include/hxString.h
@@ -39,9 +39,9 @@ public:
    #endif
 
    // Makes copy
-   static String String::create(const wchar_t *inPtr,int inLen=-1);
-   static String String::create(const char16_t *inPtr,int inLen=-1);
-   static String String::create(const char *inPtr,int inLen=-1);
+   static String create(const wchar_t *inPtr,int inLen=-1);
+   static String create(const char16_t *inPtr,int inLen=-1);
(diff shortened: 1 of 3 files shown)


hughsando:
'dup' is problematic and will be removed.  Originally, this was to convert non-gc pointers to gc-pointers.  We can't actually check whether a non-gc pointer is utf16 encoded because it will not have a header so other ways are needed.  'dup' should be replaced with 'create' calls if possible.
The 'dupConst' allocates memory outside the normal GC arenas, and this memory can't be reclaimed and so if for permanent strings. There is no reason why this can't the utf16 encoded, but I will need to check to see if its called on non-gc strings.  Some kind of 'createConst' might be better now.

hughsando:
These changes have been incorporated in recent changes that removed 'dup'.  Thanks.

PR #788, opened 2019-05-04
Build tool linear instead of quadratic relative to number of files

### LINKED ISSUE #786: [build] Parsing makefile can be slow
This step

`Parsing makefile: /some/project/mac64/haxe/cpp/Build.xml`

takes ~52 seconds in my project. Let's see if we can increase performance here.

thomasuster:
Fixes #786

When comparing build times for my project without any code changes...

Before this change: ~90 seconds
After this change: ~49 seconds

### DIFF
--- a/tools/hxcpp/BuildTool.hx
+++ b/tools/hxcpp/BuildTool.hx
@@ -369,7 +369,7 @@ class BuildTool
          //throw "No compiler defined";
       }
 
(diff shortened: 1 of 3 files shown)


thomasuster:
Blocked by https://github.com/HaxeFoundation/hxcpp/issues/791

hughsando:
This is very nice.

PR #794, opened 2019-05-13
For non-winrt 64-bit Windows builds

aprothman:
Use the 64-bit native compiler rather than the 32 to 64 cross compiler.
Changed the toolchain batch file to point relevant options and paths to amd64

### DIFF
--- a/toolchain/msvc64-setup.bat
+++ b/toolchain/msvc64-setup.bat
@@ -1,10 +1,10 @@
 setlocal enabledelayedexpansion
 @if exist "%HXCPP_MSVC%\..\..\VC\" (
-	@if not exist "%HXCPP_MSVC%\..\..\VC\bin\x86_amd64\vcvarsx86_amd64.bat" (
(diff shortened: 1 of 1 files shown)


intoxopox:
Has anyone reviewed this for consideration?

hughsando:
The scripts should be using 'host_arch' now, eg https://github.com/HaxeFoundation/hxcpp/blob/master/toolchain/msvc-arm64-setup.bat#L16
Is this PR still needed?

aprothman:
https://github.com/HaxeFoundation/hxcpp/blob/master/toolchain/msvc64-setup.bat#L3
In the Windows x64 case, the batch file that should be run is vcvars64.bat. vcvarsx86_amd64.bat sets up the 64-bit cross compiler meant to be run from 32-bit operating systems to build 64-bit applications. It forces the build process to be run from the WOW64 subsystem. The results *should* be binarily compatible, but in practice they are different. Of course, we can also run into the 32-bit process memory limits for large builds while using the cross compiler.

PR #800, opened 2019-05-24
Fix UCS2 string comparison

### LINKED ISSUE #797: Invalid string comparison
```haxe
class Main {
	static public function main() {
		var s = "𠜎zя";
		var s2 = "abя";
		trace(s > s2);
	}
}
```
That yields `false` on cpp, `true` on all other targets.

RealyUniqueName:
Fixes #797
Unblocks https://github.com/HaxeFoundation/haxe/pull/8298

### DIFF
--- a/src/String.cpp
+++ b/src/String.cpp
@@ -1540,7 +1540,6 @@ int String::compare(const ::String &inRHS) const
          }
          else
          {
-            cmp = length - inRHS.length;
             for(int i=0;i<minLen;i++)
             {
                if (__w[i]!=inRHS.__w[i])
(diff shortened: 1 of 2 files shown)


RealyUniqueName:
Fixed CI

src/String.cpp
```diff
@@ -1534,7 +1534,21 @@ int String::compare(const ::String &inRHS) const
```
Aurel300: (Cannot comment on the actual lines 1532 and 1533) There is now `String::isUTF16Encoded()`, which is an `inline` function that's basically identical to the flag check above, so `s0IsWIde` and `s1IsWide` can be defined in a more readable way.

RealyUniqueName:
> I would also drop the whitespace changes, but that's a detail.

 There is only 4 trailing whitespaces trimmed. And the rest of the file doesn't have trailing whitespaces.

PR #805, opened 2019-06-05
Attempt to fix crtbegin error (clang android)

starburst997:
For #707 to fix the error:
```
/home/joshua/Development/Android/android-ndk-r19c/platforms/android-16/arch-arm/usr/lib/../lib/crtbegin_dynamic.o:crtbegin.c:function _start_main: error: undefined reference to 'main'
```

I was able to get the DisplayingABitmap sample to run on my device using NDK 19c (both ARMV7 and ARM64) using that fix.

Worth nothing you now need to bundle the `libc++_shared.so` file inside the APK. I did it using the template folder and adding it inside the `jniLibs` folder, there is probably a more elegant way to do it (this [link](https://developer.android.com/ndk/guides/cpp-support#libc) seems to indicate that gradle can be used for that).

I was targeting platform 21.

Feel free to close it, hope this can help get official NDK 16+

### DIFF
--- a/src/hx/AndroidCompat.cpp
+++ b/src/hx/AndroidCompat.cpp
@@ -9,7 +9,7 @@
 // These functions are inlined prior to android-ndk-platform-21, which means they
 // are missing from the libc functions on those phones, and you will get link errors.
 
-#if (HXCPP_ANDROID_PLATFORM>=21) && !defined(HXCPP_ARM64)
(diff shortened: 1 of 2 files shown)


starburst997:
Merging with latest master you'll get
```
/Volumes/Android/buildbot/src/android/ndk-release-r19/bionic/libc/bionic/locale.cpp:195: error: undefined reference to '__emutls_get_address'
```

Still trying to fix that

PR #806, opened 2019-06-05
[cpp] Fix infinite array toString recursion

Aurel300:


Simn:
Please make this mergable!

PR #812, opened 2019-06-13
Fix build issues when building cppia with debugger support

waneck:


Simn:
Closing due to conflicts.

PR #817, opened 2019-06-25
Allow binding of a previous cppia class version if type hasn't been compiled

waneck:
This allows one to add partial cppia modules with only the changes
needed. This is hidden behind a compilation flag, HXCPP_CPPIA_OLD_CLASS_LINK


Simn:
Closing due to conflicts.

PR #820, opened 2019-06-26
Fix duplicate LF->CRLF encoding

Aurel300:
See HaxeFoundation/haxe#8379

 - sets `stdout` and `stderr` to binary mode (prevents `\n` from being output as `\r\n` automatically)
 - adds `HX_LF` macro, set to `\r\n` if `HX_WINDOWS` is defined, `\n` otherwise
 - replaces usages of `\n` with `HX_LF`


Aurel300:
Closed, see https://github.com/HaxeFoundation/haxe/issues/8379#issuecomment-507255656

PR #824, opened 2019-07-01
added sub and decBy to pointer

Aidan63:
Recently found myself wanting to decrement a pointer by a specific amount and discovered while there was a incBy function there wasn't a matching decBy! This adds that function to the pointer, making things just a bit cleaner that putting a negative number into incBy. Also added sub for returning a new decremented pointer, as there was an add but no matching sub.


hughsando:
Do you think it should be: 'ptr-inInt' instead of 'ptr+inInt'?

ie,  5.sub(1) should be 4 not 6?

Aidan63:
oh wow, yeah. Seems I wasn't paying attention when I made my indentation fix commit and forgot to actually have it subtract the pointers...should be fixed now though!

PR #825, opened 2019-07-04
[android] NDK20 & x86_64 support

### LINKED ISSUE #813: X86_64 support for Android
Hello!
HXCPP support X86 already, but starting August 1, 2019, Google will require 64-bit version.
Details: https://developer.android.com/distribute/best-practices/develop/64-bit
Is it possible to add X86_64 support for those who already releasing X86 version?
Thank you!

### LINKED ISSUE #823: [android] Use latest NDK
I normally use NDK 11, but while trying to implement #813 I ran into a lot of issues. Going to use this ticket as a way of trying to get the existing targets working with the latest NDK. And then see if adding x86_86 then becomes easier.

thomasuster:
Fixes #823
Fixes #813


hughsando:
I think builder can unconditionally add the x86_64 target - no need to check the define.  We do not, for example, check to see if the x86 tools are install before adding this as a potential target.

The toolchain name should remain as "android", then I think the android toolchain xml can look something like:
```xml
<setup name="androidNdk" />


<include name="toolchain/android-20+.xml" if="NDKV20+" />
<include name="toolchain/android-19-.xml" unless="NDKV20+" />

```

The existing file can then get renamed to "android-19-.xml" (and remove the 'setup androidNdk' call) and then your new file can be called android-20+.xml (also without the setup call)

Might be a good time to get rid of:
flag value="-DANDROID=ANDROID"
since HX_ANDROID is what we really want.

hughsando:
Actually, I think better names for the toolchains files would be "android-gcc.xml" and "android-clang.xml".  This way, we might use android-clang on ndk 19?  Anyhow, the critical thing is the compiler name, not the version number.

thomasuster:
Also incorporates changes from:

- https://github.com/HaxeFoundation/hxcpp/pull/827
- https://github.com/HaxeFoundation/hxcpp/pull/826 (modified)

hughsando:
Nice one, thanks!

anissen:
Can any or you please explain to me how would I go about compiling for Android using a 64 bit architecture? I am not very familiar the intricacies of building with hxcpp so please spell it out for me :)
Currently, I'm building for Android 32 bit with hxcpp through the build system set up by [luxe](https://github.com/underscorediscovery/luxe).
Thanks!

haxiomic:
@anissen for ARM64 you add `-D HXCPP_ARM64` and for desktop 64 you add `-D HXCPP_X86_64` 

In my builds I do
```hxml

-D android

--each
-D HXCPP_X86_64
--next
-D HXCPP_ARM64
--next
-D HXCPP_X86
--next
-D HXCPP_ARMV7
```

anissen:
Thanks a lot @haxiomic! 🍰

PR #826, opened 2019-07-05
NDK20 fix

codeservice:


hughsando:
Thanks for this.  I have spoken to Thomas about your change, and I think he is going to use "%p" to fix this one.

PR #827, opened 2019-07-05
More NDK20 fixes

codeservice:
This will help compile API16/armv7 on NDK20
More details here:
https://developer.android.com/ndk/guides/application_mk#app_platform
https://github.com/android-ndk/ndk/issues/358


codeservice:
Already updated in https://github.com/HaxeFoundation/hxcpp/pull/825

PR #828, opened 2019-07-07
Added a failing test demonstrating issue

thomasuster:
This pull request shows the regression mentioned in #791

### DIFF
--- a/test/haxe/TestMain.hx
+++ b/test/haxe/TestMain.hx
@@ -20,6 +20,8 @@ class TestMain {
       r.add(new TestStringHash());
       r.add(new TestObjectHash());
       r.add(new TestWeakHash());
+      r.add(new TestString());
+      
       #if cpp
       r.add(new native.TestFinalizer());
       #end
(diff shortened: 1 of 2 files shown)


hughsando:
Seems the checks are green - what version of haxe/hxcpp are you using?

thomasuster:
- Haxe Compiler 3.4.7 - (C)2005-2017 Haxe Foundation
- 
```
ed28a5b8b418f2efb3edba512022e9732a1f461e (HEAD -> thomasuster/791, origin/thomasuster/791) Added a failing test
577d88f5bf8ce8872359a584de3f4293ef491503 (upstream/master) Fix Cppia Jit floating point ++/--.  For https://github.com/HaxeFoundation/haxe/issues/8485
```

thomasuster:
Confirmed, updating to `haxe 4.0.0-rc.3` fixes the issue. Should HXCPP be patched in this case? Or should I just hop on the RC3?

hughsando:
Yes, there are problems with 3.x - it could be made to work in this case because it is the first char in the string, but chatAt(N) becomes impractical for large N with utf8 encoding and no index.  I think the solution is to move to v4.

thomasuster:
Roger willco

PR #834, opened 2019-07-19
Revert cppflag -> flag, fixes #833

### LINKED ISSUE #833: Can't compile nme fro ios
This change breaks nme compilation on ios:
from
  <flag value="-stdlib=libstdc++" if="FORCE_LIBGCC" />
  <flag value="-stdlib=libc++" if="HXCPP_CPP11" />
to
  <cppflag value="-stdlib=libstdc++" if="FORCE_LIBGCC" />
  <cppflag value="-stdlib=libc++" if="HXCPP_CPP11" />

error: #include <algorithm> not found in Rect.h

haxiomic:
See #833, recent change to `cppflag` seems to prevent these flags from being included when compiling, this PR restores them to `flag` which seems to fix compiling on iOS with `-D HXCPP_CPP11`

It's possible the reason cppflag isn't working is because `-D objc` is also set (so .mm files are generated)

### DIFF
--- a/toolchain/iphoneos-toolchain.xml
+++ b/toolchain/iphoneos-toolchain.xml
@@ -30,8 +30,8 @@
   <!-- <cppflag value="-fvisibility-inlines-hidden"/> -->
   <pchflag value="-x" />
   <pchflag value="c++-header" />
-  <cppflag value="-stdlib=libstdc++" if="FORCE_LIBGCC" />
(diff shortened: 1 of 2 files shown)


Beeblerox:
Can confirm. This change fixes compilation for iOS

hughsando:
Can you please see if the latest hxcpp changes fix this.

haxiomic:
Hey @hughsando that seems to have done the trick! Thank you :)

PR #839, opened 2019-08-02
Safe Calls

codeservice:


codeservice:
I can't use catch(Dynamic e) in CFFI integration. Dynamic object not accessible here. Only way is to use callback function.

hughsando:
Yes, catching is problematic.
What do you think of the idea of passing in a "value *" instead of the callback, then check for non-null returned exception in the c++ code?  The callback is probably going to be written in c++, not haxe, so this might be more convenient.  If it was haxe, then you could write the catching as part of the function call itself.
Adding functions to the api is a bit tricky, since neko must be considered and I have not yet added a way of checking the version of the host  (eg,  new dll, old host) but we can deal with that later.

codeservice:
I am using callback from haxe, not cpp. Passing callback haxe function as parameter to CFFI on initializtion. Because it’s on haxe side, callback code can be easily modified. 

I’ll check how it can be done from cpp.

hughsando:
Dynamic basically has a single member, mPtr, which is a hx::Object *, which is actually a 'value'.
So if you catch a Dynamic from the api implementation you can just typecast the e.mPtr to value, which should make things easier.

To do the catching in haxe, you could make a "safecaller" instead of a callback. like:
```haxe
   static function callback(e:Dynamic)  { ... }

   static function callSafe(func:Dynamic, arg1:Dynamic)
   {
      try {
         func(arg1);
      }
     catch(e:Dynamic)
     {
         callback(e);
     }
   }

   // Then, when you set the callback, do:
   library.setSafeCall(callSafe);
   // instead of
   // library.setCallback(callback);

   // from c++:
  val_call2(callSafe, func, arg1)
  // instead of
  // val_call_safe1(func,arg1,callback)
```

codeservice:
Using haxe safecaller I can't catch other cpp exceptions, only haxe Dynamic type.
static function callSafe(func:Dynamic, arg1:Dynamic)
   {
      try {
         func(arg1);
      }
     catch(e:Dynamic) //limited to haxe exception type
     {
         callback(e);
     }
   }

In cpp I can use catch(...) for any type, making call really safe.
I'll try this option "passing in a "value *" instead of the callback"

codeservice:
I think I am wrong, its possible if I wrap val_call in try / catch:
try {
val_call2(callSafe, func, arg1)
} catch(...){
//...
}

For me performance important, need to check haxe safecaller version ...

hughsando:
This should work if you static-link the nme library.  I'm not sure if the
type-information is consistent between dlls and the main program - but I
guess if the catch does not depend on the type, it might just work.

codeservice:
Its for different library, not NME. In some cases I can't statically link other libraries. For example, I didn't find way statically link LuaJIT, don't think this even possible. Probably easiest way for me to keep using custom hxcpp modification from my local hxcpp version.

PR #841, opened 2019-08-02
pcre-8.43

codeservice:


Uzume:
This is nice but shouldn't we be moving to PCRE2? Along with the rest of Haxe, see HaxeFoundation/haxe#10491.

Simn:
Agreed, I think we can close this.

PR #855, opened 2019-10-04
Added Android's x86_64 architecture for older NDKs

lehonma:
Added x86_64 support in the GCC toolchain for older NDKs - I tested with NDK15c, but should be compatible with 14b as well.


tobil4sk:
@Simn This would also be useful to merge, as ndk r15c is what lime uses for its android build and without this, an attempt to create an x86_64 build just quietly gives a 32 bit build instead.

tobil4sk:
Many thanks!

PR #857, opened 2019-10-11
Add write barriers for Anons

RobDangerous:
Usual disclaimer - I don't really know what I'm doing. Also didn't have any actual problems here.

### DIFF
--- a/include/hx/Anon.h
+++ b/include/hx/Anon.h
@@ -80,6 +80,12 @@ public:
       fixed->hash = inName.hash();
       fixed->key = inName;
       fixed->value = inValue;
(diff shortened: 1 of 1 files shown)


hughsando:
Yes, this seems pretty sensible.  You won't see the 'setFixed' code without  a inline anon-object definition (rather than loading from file), and the other code won't show up without @:fixed metadata.  I think I still need to do something in the @:fixed constructor case.

RobDangerous:
Lots of setFixed in my code but I didn't even know about @:fixed yet. Might come in handy, thanks!

PR #858, opened 2019-10-12
Add write barriers to fixed anon constructors

RobDangerous:
And here is yet another blind fix based on your comment to the previous pull request.

PS: Just now also added a fix for the previous pull request.

### DIFF
--- a/include/hx/Anon.h
+++ b/include/hx/Anon.h
@@ -83,6 +83,9 @@ public:
       if (inValue.type == ::cpp::Variant::typeObject) {
         HX_OBJ_WB_GET(this, inValue.valObject);
       }
+      else if (inValue.type == ::cpp::Variant::typeString) {
+        HX_OBJ_WB_GET(this, inValue.valStringPtr);
+      }
       return this;
    }
    inline VariantKey *getFixed()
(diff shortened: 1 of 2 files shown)


hughsando:
This is looking good, but you do not need write-barriers on the names of 'setFixed' or the AnonStructs because these are known to be generated from const strings allocated outside the GC system (hence no entries in the __Mark code too).  It should still work, but with a tiny performance overhead.

RobDangerous:
Okidoki, removed the stringy write barriers.

include/hx/Anon.h
```diff
@@ -81,10 +81,7 @@ class HXCPP_EXTERN_CLASS_ATTRIBUTES Anon_obj : public hx::Object
       fixed->key = inName;
       fixed->value = inValue;
       if (inValue.type == ::cpp::Variant::typeObject) {
-        HX_OBJ_WB_GET(this, hx::PointerOf(inValue.valObject));
-      }
-      else if (inValue.type == ::cpp::Variant::typeString) {
```
hughsando: Actually, this one should stay.  The value could be a variable, while the name is known to be constant.

RobDangerous: Fixed.

hughsando:
Brilliant.  Thanks for the changes - no good deed goes unpunished.

RobDangerous:
Thanks for all the help in getting this working!

PR #861, opened 2019-10-29
Fix lost field name when allocating MySQL results

jonasmalacofilho:
Fixes: e11b94aae518 ("More utf16 fixes")
Fixes: HaxeFoundation/record-macros#43


emugel:
I applied your patch and it indeed solves the ""Missing key xx"" problem with `record-macros`.

jonasmalacofilho:
I guess I should also try to enable the MySQL tests...

:cold_sweat:

jonasmalacofilho:
Just when I was about to celebrate a first – configuring new stuff in CI without messing up a couple of times – a wild network problem appears:

```
$ haxelib install compile-cpp.hxml --always
Loading info about the required libraries
Error: Failed with error: Http Error #503
```

---

When things go back to normal I can force-push to trigger a rebuild.

Issue #859, opened 2019-10-14
Sqlite cnx.request return error!


sonygod:
Sqlite cnx.request return error!

same code test between hashlink and hxcpp

hashlink return correct results and hxcpp throw error.

        import sys.db.Sqlite;

        class CustomTrace {
            public static function init() {
                var oldTrace = haxe.Log.trace;
                haxe.Log.trace = function(v:Dynamic, ?pos:haxe.PosInfos) {
                    oldTrace("[" + Date.now().toString() + "] " + Std.string(v), pos);
                };
            }
        }

        typedef Scroe = {
            var id:Int;
            var roomID:Int;
            var round:Int;
            var cards:Array<Int>;

            var earnMoney:Int;
            var loseMoney:Int;
            var fk:Int;
            var time:Date;
            var key:Int;
        }

        class TestSqlite {
            public static function main() {
                CustomTrace.init();
                var cnx = Sqlite.open("mybase.db");
                cnx.request("
                CREATE TABLE IF NOT EXISTS fa_game (
                    id integer NOT NULL PRIMARY KEY AUTOINCREMENT,
                    us_id integer  NOT NULL DEFAULT 0,
                    room_id integer  NOT NULL DEFAULT 1,
                    round integer NOT NULL DEFAULT 0,
                    coeds TEXT NOT NULL,
                    win integer NOT NULL DEFAULT 0,
                    lost integer NOT NULL DEFAULT 0,
                    fk integer NOT NULL DEFAULT 0,
                    time DATETIME NULL DEFAULT CURRENT_TIMESTAMP,
                    times integer  NOT NULL,
                    score integer NOT NULL DEFAULT 0,
                    money integer NOT NULL DEFAULT 0,
                    isBanker integer NOT NULL DEFAULT 0
                )
                ");

                var allScores:Array<Scroe> = [];

                for (i in 0...4) {
                    allScores.push({
                        id: i,
                        roomID: i,
                        fk: i,
                        round: i,
                        key: i,
                        time: Date.now(),
                        cards: [i, i + 1, i + 2],
                        loseMoney: i,
                        earnMoney: 1
                    });
                }

                trace(allScores.length);
                var arr = [];
                for (p in allScores) {
[...]

PR #862, opened 2019-11-04
Handle .length and .nfields on empty result sets

jonasmalacofilho:
Forward the number of changed rows for statements other than SELECT.
This prevents an exception to be thrown with SQLite,[1] and also matches
the corresponding Neko documentation[2].

Both hx_sqlite_result_get_length and _hx_mysql_result_get_nfields are
changed to work in these scenarios.  The latter change fixes a different
yet related error with MySQL, where .nfields would raise a similar
exception.

Finally, this also fixes an issue with running the MySQL tests locally,
as the 'hxcpp' test database will be reused and has to be cleared.

[1] Sqlite cnx.request return error! (hxcpp#859)
[2] Neko SQLite API (https://nekovm.org/doc/view/sqlite/)

Fixes: #859

### DIFF
--- a/src/hx/libs/mysql/Mysql.cpp
+++ b/src/hx/libs/mysql/Mysql.cpp
@@ -196,6 +196,9 @@ int  _hx_mysql_result_get_length(Dynamic handle)
 **/
 int  _hx_mysql_result_get_nfields(Dynamic handle)
 {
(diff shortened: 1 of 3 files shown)


sonygod:
+1

PR #873, opened 2019-11-29
Fix "cannot pass object of non-trivial type" errors in clang

RobDangerous:
clang doesn't like all of the PointerOf calls that might be happening.
There's one of those problems with Arrays at Kode/Kha#1168 and I changed PointerOf(Dynamic) to PointerOf(ObjectPtr) to catch it.
@nadako ran into the same problem with TNonGcStringSet as well so I added that too and just return 0. Not sure about that one though.

### DIFF
--- a/include/Array.h
+++ b/include/Array.h
@@ -415,8 +415,11 @@ template<typename TYPE> inline bool ContainsPointers()
    return TypeContainsPointers( (TYPE *)0 );
 }
 
-inline const void *PointerOf(Dynamic &d) { return d.mPtr; }
+struct TNonGcStringSet;
+
+template<typename T> inline const void *PointerOf(hx::ObjectPtr<T> &o) { return o.mPtr; }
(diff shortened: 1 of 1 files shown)


hughsando:
This is a bit unfortunate - the real ("easy") fix is for the haxe compiler to generate these structs with with 'Dynamic' template type, rather than the specific object types - there should be no speed loss because the access to these members is Dynamic anyhow.  This is how Array works and why its 'pointerOf' is so simple.
Given that updating the haxe compiler is a long-term proposition, I think we need a new template arrangement in GcTypeInference, 'GcPointerOf' to extract the pointers.  Using Dynamic or ObjectPtr&lt;T> is not really enough since types will get missed (Dynamic does not match ObjectPtr).  The full list of classes should be like MarkMember in GcTypeInference.h although perhaps Variant is not required.

Simn:
This has conflicts and there was no follow-up, so I'll assume that this can be closed.

PR #880, opened 2020-01-23
Uses "%ls" if isUTF16Encoded for print/trace

### LINKED ISSUE #842: how to support utf-8 for trace ?
how to support utf-8 for trace ?
-D What?

R32:
resolve #842 

```haxe
class Main {
    static function main() {
        var s = "Akuma 豪鬼";
        trace(s);
        Sys.println(s);
    }
}
```
output:

```bash
src/Main.hx:6: Akuma 豪鬼
Akuma 豪鬼
```

### DIFF
--- a/src/hx/StdLibs.cpp
+++ b/src/hx/StdLibs.cpp
@@ -288,7 +288,14 @@ void __trace(Dynamic inObj, Dynamic info)
    hx::strbuf convertBuf;
    if (info==null())
    {
+#if defined(HX_SMART_STRINGS) && defined(HX_WINDOWS)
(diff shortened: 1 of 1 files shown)


sonygod:
Very nice

sonygod:
@R32  your branch may get some error with emcc target.

https://github.com/HaxeFoundation/hxcpp/issues/883

tobil4sk:
This doesn't work in the general case, I think it only works on system where a regional locale is set that already supports these characters. Usually the locale is set to something that only supports a specific language or small set of languages so this will still give garbled output for characters outside the locale range.

PR #881, opened 2020-01-27
__hxcpp_parse_int now skips leading space chars

### LINKED ISSUE #869: `__hxcpp_parse_int` can't parse hexadecimal with leading whitespaces
```haxe
trace(Std.parseInt('  -0x10')); // 0
```
Related Haxe issue: https://github.com/HaxeFoundation/haxe/issues/8970

Aidan63:
Fixes #869. Uses `isstring` to catch the most common space characters, also added a few tests in the std section.


Simn:
I didn't see this PR and now I've caused a merge conflict with https://github.com/HaxeFoundation/hxcpp/commit/cb38fcc2f89d59c44baa13c10af3ebe15fe1e353

We should look into merging this because, among other things, the current `str[1]` could actually cause out-of-bounds access, which this PR addresses as well. Could I trouble you to resolve the conflict and incorporate the change from my commit?

Aurel300:
(For the record it does not cause out-of-bounds access. The second part of that check is only executed if `str[0]=='0'` because of short-circuiting, and a null-terminated string has to have its null-byte accessible.)

Simn:
You're right, I somehow had it committed to memory that these strings weren't null-terminated, but they would obviously have to be given that the C string functions are being used there.

Aidan63:
I've just merged master into this branch and added some extra tests for parsing negative hex literals prefixed with whitespace.

Simn:
Thank you!

Unfortunately, our CI is being difficult, but I'll merge this if the windows tests come up green.

Aidan63:
botched a bit of the merge which broke some haxe tests, but they're all passing for windows now.

Simn:
Thank you!

PR #891, opened 2020-03-10
Add keyValueIterator to arrays

### LINKED ISSUE #871: Array.keyValueIterator() implementation
See https://github.com/HaxeFoundation/haxe/pull/7422

Aurel300:
Closes #871 (and unblocks https://github.com/HaxeFoundation/haxe/pull/7422).

I mostly followed the implementation of `iterator` for Arrays. Some things that I am not sure about:

 - `include/hx/Object.h` has a `clsIdArrayIterator` - should there ba a `clsIdArrayKeyValueIterator`? I couldn't figure out what these `clsId` things actually do. For now the KV iterator is also marked as `clsIdArrayIterator`.
 - `src/cppia/Cppia.h`, line 116 - I added `afKeyValueIterator` to the array functions right after the existing `afIterator`. Is there an external dependency on the exact order of these values?
 - `src/cppia/ArrayBuiltin.cpp`, line 1697 - is the conversion to `destType` there just "convert to object"?
 - `include/Array.h` and `src/Array.cpp` - I added a `hx::ArrayKeyValueIterator` class to mirror the `hx::ArrayIterator` one. I could not define the `Dynamic next()` method in the header file because I cannot include `Anon.h` in `Array.h` (cyclic dependency). The implementation of is in `src/Array.cpp`.

The CI failure seems to be related to the last point. I guess `src/Array.cpp` (or its object file) is not being linked wherever things from `src/Array.h` are used?

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -10,6 +10,7 @@
 4.0.19
 ------------------------------------------------------------
 
+* Add Array.keyValueIterator
 * General Utf16 string improvements
 * Limit the amount of recursion in toString function
 * Add float32 support to cppia
--- a/include/Array.h
+++ b/include/Array.h
@@ -93,6 +93,34 @@ public:
    Array<FROM> mArray;
 };
 
+// --- ArrayKeyValueIterator -------------------------------------------
+template<typename FROM,typename TO>
+class ArrayKeyValueIterator : public cpp::FastIterator_obj<Dynamic>
+{
+public:
+   HX_IS_INSTANCE_OF enum { _hx_ClassId = hx::clsIdArrayIterator };
+
+   ArrayKeyValueIterator(Array<FROM> inArray) : mArray(inArray), mIdx(0) { }
+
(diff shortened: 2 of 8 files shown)


hughsando:
The clsId is for "Std.is" - since you can't ask this for iterators (?), it does not matter.

The order of the ArrayFunc enum needs to match the gArrayFuncNames and gArrayArgCount order, but these are not saved to disk anywhere so just need to be consistent for the single compile.

It is quite possible the destType will be String (eg, "trace").  With some Dynamic stuff, it might end up being Int too, but the result would not be important here, since it's undefined behaviour.

I can't see the error, but I think it's because the c++ code needs to be able to "see" the definition if its templated.  The best place to put this is in "include/hx/Operators.h" which comes after everything else is included.

hughsando:
but also, good work.

nadako:
@Aurel300 let's do this for 4.1 :)

PR #892, opened 2020-03-14
Map toString brackets

RblSb:
Changes brackets style in maps for https://github.com/HaxeFoundation/haxe/pull/9260

### DIFF
--- a/src/hx/Hash.cpp
+++ b/src/hx/Hash.cpp
@@ -69,7 +69,7 @@ void __int_hash_set(HX_MAP_THIS_ARG,int inKey,const Dynamic &value)
          }
          else if (type==vtFloat)
          {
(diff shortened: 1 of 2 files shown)


hughsando:
Is the intention here to have Map display differently from Dynamic ?
If you want to restrict it to a particular version (eg, haxe 4.1) there is going to need to be some conditional-compiling in the c++ code.

RealyUniqueName:
The intention is to have Map display the same as in Haxe syntax: `[key => value]`

Simn:
Hmm yeah, updating hxcpp should not cause any printing changes on Haxe versions < 4.3.

Simn:
This looks good to me now that everything is hidden behind an API check.

PR #902, opened 2020-05-08
Invoke cppia without script name in arguments

### LINKED ISSUE #901: [Cppia] Sys.args contain name of program file
When running `haxelib run hxcpp myProgram.cppia -v` or `bin/Cppia myProgram.cppia -v` calling `Sys.args ()` will return `["myProgram.cppia", "-v"]` instead of just `["-v"]` (like e.g. a cpp compiled version does). 
According to https://api.haxe.org/Sys.html argument list should not contain interpreter nor name of program file.

Aurel300:
Closes #901

### DIFF
--- a/tools/hxcpp/BuildTool.hx
+++ b/tools/hxcpp/BuildTool.hx
@@ -1540,7 +1540,7 @@ class BuildTool
             exe = '"$exe"';
          #end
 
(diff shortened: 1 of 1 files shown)


Aurel300:
Actually never mind, this will just break cppia. 🤦

PR #905, opened 2020-05-11
Pass the byte length into sqlite requests instead of char length (Fixes #900)

### LINKED ISSUE #900: [sqlite] not support utf8.
class Main {
		static function main() {
			var cnx = sys.db.Sqlite.open("./backup.data");
			trace('create table now');
			cnx.request("
					CREATE TABLE IF NOT EXISTS fa_customer (
					id integer NOT NULL PRIMARY KEY AUTOINCREMENT,
					
					nickname text(255)
					
					
					
					)");

			trace('create table finish');

			var sql = 'INSERT INTO fa_customer (nickname) VALUES (${cnx.quote("豪鬼")})';
			trace(sql);
			cnx.request(sql);

			trace("sqlite init finish");
		}
	}

build.hxml

    -cp src
    -D analyzer-optimize
    -main Main
    --each
    --hl test.hl
    --next
    --cpp bin

cpp version

I:\test\testcpp\bin>Main.exe
src/Main.hx:5: create table now
src/Main.hx:16: create table finish
src/Main.hx:19: INSERT INTO fa_customer (nickname) VALUES ('璞')
Error : ValueException

hl verion

I:\test\testcpp>hl test.hl
src/Main.hx:5: create table now
src/Main.hx:16: create table finish
src/Main.hx:19: INSERT INTO fa_customer (nickname) VALUES ('豪鬼')
src/Main.hx:22: sqlite init finish

Aidan63:
Didn't seem to be an existing way to do it so I added a new utf8 string conversion function which allows retrieving the byte length from `TConvertToUTF8`.

20 will now be passed into the `sqlit3_prepare` function for `SELECT '豪鬼' test` instead of 16. The reduced example from #900 now outputs `{ test => è±ªé¬¼ }` instead of an error being thrown.

Was going to add some tests but there don't seem to be any existing sqlite tests in the main haxe repo or hxcpp.

### DIFF
--- a/include/hxString.h
+++ b/include/hxString.h
@@ -154,8 +154,7 @@ public:
 
    inline const char *&raw_ref() { return __s; }
    inline const char *raw_ptr() const { return __s; }
-   const char *utf8_str(hx::IStringAlloc *inBuffer = 0,bool throwInvalid=true) const;
-   const char *utf8_str_len(hx::IStringAlloc *inBuffer = 0,int *byteLength = 0,bool throwInvalid=true) const;
+   const char *utf8_str(hx::IStringAlloc *inBuffer = 0,bool throwInvalid=true, int *byteLength = 0) const;
    const char *ascii_substr(hx::IStringAlloc *inBuffer,int start, int length) const;
    inline const char *c_str() const { return utf8_str(); }
    inline const char *out_str(hx::IStringAlloc *inBuffer = 0) const { return utf8_str(inBuffer,false); }
--- a/include/hxString.h
+++ b/include/hxString.h
@@ -155,6 +155,7 @@ public:
    inline const char *&raw_ref() { return __s; }
    inline const char *raw_ptr() const { return __s; }
    const char *utf8_str(hx::IStringAlloc *inBuffer = 0,bool throwInvalid=true) const;
+   const char *utf8_str_len(hx::IStringAlloc *inBuffer = 0,int *byteLength = 0,bool throwInvalid=true) const;
    const char *ascii_substr(hx::IStringAlloc *inBuffer,int start, int length) const;
    inline const char *c_str() const { return utf8_str(); }
    inline const char *out_str(hx::IStringAlloc *inBuffer = 0) const { return utf8_str(inBuffer,false); }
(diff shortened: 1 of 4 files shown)


src/String.cpp
```diff
@@ -1573,6 +1573,24 @@ const char * String::utf8_str(hx::IStringAlloc *inBuffer,bool throwInvalid) cons
...
+#ifdef HX_SMART_STRINGS
+   if (isUTF16Encoded())
+   {
+      return TConvertToUTF8(__w, byteLength, inBuffer, throwInvalid);
+   }
+   else
+   {
+      *byteLength = length;
+      return __s;
+   }
```
jonasmalacofilho, line 1587: I don't think this `else` block is necessary.

src/String.cpp
```diff
@@ -1573,6 +1573,24 @@ const char * String::utf8_str(hx::IStringAlloc *inBuffer,bool throwInvalid) cons
...
+#else
+   *byteLength = length;
```
jonasmalacofilho, line 1589: It might be wise to keep consistency with `TConvertToUTF8` and check that `byteLength` isn't NULL.

include/hxString.h
```diff
@@ -155,6 +155,7 @@ class HXCPP_EXTERN_CLASS_ATTRIBUTES String
    inline const char *&raw_ref() { return __s; }
    inline const char *raw_ptr() const { return __s; }
    const char *utf8_str(hx::IStringAlloc *inBuffer = 0,bool throwInvalid=true) const;
+   const char *utf8_str_len(hx::IStringAlloc *inBuffer = 0,int *byteLength = 0,bool throwInvalid=true) const;
```
jonasmalacofilho, line 158: I'm terrible at C++: is overloading `utf8_str` an option here?

src/String.cpp
Aidan63: Yeah, must have been overthinking it, should be gone now

src/String.cpp
Aidan63: Good catch, should now be wrapped in a nullptr if check.

include/hxString.h
Aidan63: Wasn't sure if it would be better to add another optional argument or to make another function. Thinking about it some more I think the extra argument is better, so `utf8_str_len` should be gone now.

Aidan63:
Yeah, I'm on windows so that's probably it. Thanks for point out the test location, added a quick extra request to the sqlite section, lets see what the ci thinks!

sonygod:
well. seem fixed now.

hughsando:
Thanks!

PR #908, opened 2020-05-21
Fix issue with nullpointer when debugging 

m0rkeulv:
This change solves an issue i have when using VS code debugger, i am not sure why an empty string is passed to lookupFileName , but checking if the string is empty before trying to do a compare solves my nullpointer crash/problem.

### DIFF
--- a/src/hx/Debugger.cpp
+++ b/src/hx/Debugger.cpp
@@ -875,9 +875,11 @@ private:
    // when evaluating breakpoints
    static const char *LookupFileName(String fileName)
    {
+      if (fileName.length == 0) return 0;
+       
       for (const char **ptr = hx::__hxcpp_all_files; *ptr; ptr++)
       {
-         if (fileName.length>0  && !strcmp(*ptr, fileName))
+         if (!strcmp(*ptr, fileName))
             return *ptr;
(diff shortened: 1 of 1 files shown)


m0rkeulv:
It might also be possible to do the following for more performant code.
```cpp
 static const char *LookupFileName(String fileName)
   {
      if (fileName.length>0) return 0;

      for (const char **ptr = hx::__hxcpp_all_files; *ptr; ptr++)
      {
         if (!strcmp(*ptr, fileName))
            return *ptr;
      }
....
```
Or someone look into why /how we get an empty String in the first place, which is probably the best solution.

hughsando:
Yes, I'm not sure why the debugger code is passing in a null string here - possibly a parsing issue?  This could be coming from outside hxcpp, so it makes sense to check for null.
I think your second solution is better because it also catches the "scriptable" case.  It might even be worth logging something to the console the first time null is encountered to let the debug users know something is not right.

m0rkeulv:
@hughsando 
i have changed to code to the second solution,  not sure why  some of the builds fails though.

hughsando:
Seems it should only make things better, so lets do it.

Issue #910, opened 2020-05-29
Checking thread equals crash


codeservice:
Rare issue, observed on ARM64 android, haxe3 and haxe4. (armv7 fine)
When I compare threads (both not null):  thread.equals(Thread.current()) android can silently crash from time to time. Error usualy looks like: A/libc: Fatal signal 11 (SIGSEGV), code 128, fault addr 0x0 in tid 2412 (e.app.testapp)

PR #916, opened 2020-06-05
Fix for thread issue 910

codeservice:
Because I have this issue on haxe 3 too, needed to look what went wrong here https://github.com/HaxeFoundation/hxcpp/issues/910
I think Thread current can return wrong thread. I found fix that helping in my case. Not sure if this actually fixing in all cases. For example when thread run from thread, etc.
Problem can be related to declaring tlsCurrentThread as global var.

### DIFF
--- a/src/hx/Thread.cpp
+++ b/src/hx/Thread.cpp
@@ -237,7 +237,6 @@ THREAD_FUNC_TYPE hxThreadFunc( void *inInfo )
    info[0] = (hxThreadInfo *)inInfo;
    info[1] = 0;
 
-	hxThreadInfo *oldinfo = tlsCurrentThread;
 	tlsCurrentThread = info[0];
 
 	hx::SetTopOfStack((int *)&info[1], true);
@@ -259,8 +258,7 @@ THREAD_FUNC_TYPE hxThreadFunc( void *inInfo )
 
 	hx::UnregisterCurrentThread();
 
-	//tlsCurrentThread = 0;
-	tlsCurrentThread = oldinfo;
+	tlsCurrentThread = 0;
 
 	THREAD_FUNC_RET
 }
--- a/src/hx/Thread.cpp
+++ b/src/hx/Thread.cpp
@@ -237,6 +237,7 @@ THREAD_FUNC_TYPE hxThreadFunc( void *inInfo )
    info[0] = (hxThreadInfo *)inInfo;
    info[1] = 0;
 
+	hxThreadInfo *oldinfo = tlsCurrentThread;
 	tlsCurrentThread = info[0];
 
 	hx::SetTopOfStack((int *)&info[1], true);
@@ -258,7 +259,8 @@ THREAD_FUNC_TYPE hxThreadFunc( void *inInfo )
 
 	hx::UnregisterCurrentThread();
 
-	tlsCurrentThread = 0;
+	//tlsCurrentThread = 0;
+	tlsCurrentThread = oldinfo;
 
 	THREAD_FUNC_RET
 }
@@ -302,10 +304,13 @@ static hxThreadInfo *GetCurrentInfo(bool createNew = true)
 	if (!info && createNew)
 	{
 		// Hmm, must be the "main" thead...
-		info = new hxThreadInfo(null(), 0);
-		sMainThreadInfo = info;
-		hx::GCAddRoot(&sMainThreadInfo);
-		tlsCurrentThread = info;
+		if(!sMainThreadInfo) {
+		  info = new hxThreadInfo(null(), 0);
+		  sMainThreadInfo = info;
+		  hx::GCAddRoot(&sMainThreadInfo);
+		  tlsCurrentThread = info;
+                }else
+		  info = (hxThreadInfo *)sMainThreadInfo; 			
 	}
 	return info;
 }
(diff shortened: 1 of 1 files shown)


src/hx/Thread.cpp
```diff
@@ -258,7 +259,8 @@ THREAD_FUNC_TYPE hxThreadFunc( void *inInfo )
 
 	hx::UnregisterCurrentThread();
 
-	tlsCurrentThread = 0;
+	//tlsCurrentThread = 0;
```
Aurel300, line 262:
```suggestion

```

src/hx/Thread.cpp
```diff
@@ -302,10 +304,13 @@ static hxThreadInfo *GetCurrentInfo(bool createNew = true)
...
-		sMainThreadInfo = info;
-		hx::GCAddRoot(&sMainThreadInfo);
-		tlsCurrentThread = info;
+		if(!sMainThreadInfo) {
+		  info = new hxThreadInfo(null(), 0);
+		  sMainThreadInfo = info;
+		  hx::GCAddRoot(&sMainThreadInfo);
+		  tlsCurrentThread = info;
+                }else
+		  info = (hxThreadInfo *)sMainThreadInfo;
```
Aurel300, line 313:
```suggestion
		if (!sMainThreadInfo) {
		  info = new hxThreadInfo(null(), 0);
		  sMainThreadInfo = info;
		  hx::GCAddRoot(&sMainThreadInfo);
		  tlsCurrentThread = info;
                } else {
		  info = (hxThreadInfo *)sMainThreadInfo; 			
                }
```

Aurel300:
I don't know enough about this part of the code to see why it would be correct. Why are you adding a root but not removing it?

codeservice:
It’s part of original code. I think main thread only one for whole app and doesn’t require removing from root. Question do we need place each thread to root and remove after running, I don’t see this in thread code. For me thread looks like require some attention.

hughsando:
`tlsCurrentThread` is not really a global - it has a different value depending on the identity of the native thread that is querying the value.  So, if tlsCurrentThread is null, and sMainThread is not null, then the native thread is not the main thread.
Roots are not needed for the non-main thread because their memory location defines the top-of-stack and will always be hit by the conservative collector.

The threadinfo is indirectly created by the haxe in the Thread.create function so this suggests to me that you are querying the  info from a thread created by a non-haxe thread call (foreign thread).

I have not really though about this, but if foreign threads don't get an info, then your compare is not going to work.   Also, you don't get things like "sendMessage" which requires some resources.    It is not clear to me how these could be properly managed from foreign threads.

All attached threads do get a stack context, but these are recycled as threads attach and detach so their identity as a thread affinity is not well defined.

So I really want to say something like "only threads created from haxe can use sendMessage, TLS, and return a valid value for 'current'".
This does not help you I guess.  Maybe a "getNativeThreadId" is what your really need if you want to compare thread identity.

codeservice:
I am using only haxe threads. I had this rare issue with Thread.current on arm64 only, maybe can be repeated on other platforms. So for some reason calling Thread.current many times can return not main thread in my code, and this code has only main thread. I have Thread.current check for different logic where I need to prevent some events from other then main thread. And this check can crash some time.

" if tlsCurrentThread is null, and sMainThread is not null, then the native thread is not the main thread." Does this mean android arm64 can run main code in different thread at random time?

hughsando:
Android certainly can. With an app, there are usually at least 2 threads involved - the UI thread, and the render thread.  Possibly all system callbacks will happen on other threads.  Now, maybe the main/init code is called at startup from the ui thread, but the majority of the calls happen from the render thread - which is logically the "main thread".  These startup and render phases do not overlap, but will have different native thread ids.  So maybe you get the bug depending on whether the check the "current" on the startup code.
If you are building an command-line exe with no system ui threads then the situation should be the same as a windows command line exe, and not have this issue.

codeservice:
In my fix, I am always returning main thread when current null and main is not null. Original Hxcpp code creates another main thread and place it in root, is this correct behavior? I am asking because operations, like thread “equals” in my case can trigger silent crash. Maybe hxcpp need to throw error in this case? Something like, “current thread unreachable”

hughsando:
I think the correct answer is to assume it is a new foreign thread (ie, an android render thread) and create a new info structure and store it in the tls and a gcRoot.  I probably then need to listen for the native thread destruction and release the info at this stage, although I'm not 100% sure I can do this.
The alternative would be to create a "light" version of the structure that is not stored as a root (ie, collected) but just has access to the ThreadNumber, which should allow comparisons, but not things like sendMessage.

codeservice:
It doesn't sound like simple task. I can tell, in my code I was able to see equals works some time in threadless haxe app (main thread only). Equals result was true, false or unexpected crash. Results looks random, its really hard to fix anything without clean, repeatable issue example. I will continue looking for this problem too. For now I am going to always return main thread, this prevent my app from crash and seems didn't break anything else. For me good enough to assume, there is only one main thread, because haxe code can operate with this thread only, for now.

hughsando:
This should be fixed via https://github.com/HaxeFoundation/hxcpp/commit/95a4e943b9516be8846bd5846b24cbf73b2754e3

PR #921, opened 2020-06-11
#1 + #2 + #3 fixes and improvements pr pack

### LINKED ISSUE #1: Full file path in debugger
Can you print absolute path in debugger stack trace output.
It's important when using openfl, for example.
in stack trace i haxe lines like "123 line in flash/Lib.hx", but there are few Lib.hx files (one in haxe std lib, one in openf, one in openfl-native) and openfl-native not even in class path.

sibest:
This is the first part (1/3) of pull requests I will publish today. 

This one just fixes __SetSizeExact which causes an invalid entry put into the nursery when the realloc size is 0.

### DIFF
--- a/include/Array.h
+++ b/include/Array.h
@@ -450,6 +450,8 @@ template<typename TYPE> inline bool ContainsPointers()
 
 inline const void *PointerOf(Dynamic &d) { return d.mPtr; }
 inline const void *PointerOf(String &s) { return s.raw_ptr(); }
+template<typename T>
+inline const void *PointerOf(::Array<T> &a) { return a.mPtr; }
 inline const void *PointerOf(...) { return 0; }
 
 
--- a/include/hx/GC.h
+++ b/include/hx/GC.h
@@ -10,9 +10,89 @@
 //  the end of the data.
 // When HX_SMART_STRINGS is active, a bit says whether it is char16_t encoded.
 
+
+// BENCHMARK: To-be tested
+// #define IMMIX_USE_QUICKSEARCH_FOR_BLOCKS
+
+// #define IMMIX_STRESS_TEST_CONSERVATIVE
+
+// Allocate this 2^many blocks at a time - this will increase memory usage % when rounding to block size must be done.
+// However, a bigger number makes it harder to release blocks due to pinning
+#define IMMIX_BLOCK_GROUP_BITS  5
+
+#ifdef HXCPP_GC_BIG_BLOCKS
+   #define IMMIX_BLOCK_BITS      16
+   typedef unsigned int BlockIdType;
+#else
+   #define IMMIX_BLOCK_BITS      15
+   typedef unsigned short BlockIdType;
+#endif
+
+#ifdef  HXCPP_GC_SHORT_ROWS
+// Each line ast 64 bytes (2^6)
+#define IMMIX_LINE_BITS    6
+// This is necessary to allow IMMIX_LINE_BITS < 7
+#define IMMIX_LOOP_COUNT_ROWS  
+#else
+// Each line ast 128 bytes (2^7)
+#define IMMIX_LINE_BITS    7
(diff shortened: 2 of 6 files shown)


hughsando:
I think `alloc(0)` should return non-null in generational case, like in non-generational case.
The easiest solution here would be to use something like `if (size<4) size=4` in the nursery.

sibest:
why ?

hughsando:
Consistency mainly.  I also can't be sure that some application like OpenFL is not calling alloc, and checking the result against NULL to see if it worked and throwing a error if it did not.

sibest:
Ok, I will change it, but it will waste space in the memory blocks.

sibest:
Well actually at the current state if anything would ask for a 0-byte realloc or alloc, it would crash because of the size check in GetEnclosingNurseryType (this is the reason of the fix), and returning 0 seems fine for now.

sibest:
I've uploaded the #2 fixes.

The HX_USE_INLINE_IMMIX_OPERATOR_NEW was flawed, yes it was fast, but it could lead to race conditions:

`inline static void *alloc`
doesn't have any thread lock, and while a thread was collecting it could modify a block leading
to a crash.

**What I've done:**
- I've rewritten the management of thread locks in Immix.cpp to mantain the inliner speed:
Now when a LocalAllocator acquires ownership of a block, it is mantained until the allocator releases it. When the block is released it must be marked before it can used again. 
- Fixed countRows (it was wrongly using IMMIX_LINES instead of IMMIX_USEFUL_LINES)
- Fixed memory statistics: While debugging my game, I noticed that sometime the MEMORY_FRAGMENTATION reported very wrong values (For example 1920GB in use), it was because spaceOversize was set to 0 when a collection did start, and *mFraggedRows += (spaceOversize - spaceFirst)>>IMMIX_LINE_BITS would be a negative number
- Fixed Array.h, added hx::PointerOf Array

hughsando:
It will only crash if nursery is enabled, which may not be tested.  So I
don't want the case where someone enables the nursery and their code stops
working.
As for arrays, I think that it is better for SetSizeExact(0) to set the
mBase to null and release the previous value - perhaps just as a special
case.

hughsando:
The state for "inline static void alloc" is maintained in ImmixAllocator *alloc, which is unique per thread, so there should be no obvious race conditions based simply on it being static.  It should indeed have complete and unique ownership of any block assigned to it.
All local allocators should either be in "gc free zones" (ie, not mutating haxe data) or blocking in the collection code when a collection is performed.  Foreign threads if not attached properly can cause problems.

sibest:
The race condition is there, because you still change the BlockData while maybe a reclaim is in progress. I've tested it deeply.

sibest:
The last commit is the part #3, just some improvements (at least for me)
- I've added the ability to use short rows (64bytes) instead of 128bytes to reduce fragmentation
it can be enabled from the project by the config

`<haxedef name="HXCPP_GC_SHORT_ROWS" />`

- I've added a simple hash vector to search inside the memory pool, it can be enabled for blocks too
with the define `#define IMMIX_USE_QUICKSEARCH_FOR_BLOCKS` in GC.h but i didn't benchmark it yet (currently it's using a binary search)

hughsando:
The reclaim on a block should only touch the non-marked rows, and the haxe code should only touch the marked rows, or the rows that were designated holes after the reclaim has finished.  If this is being violated, then it needs to be addressed once the reason is found.
If you have any example code that shows this issue, it would be helpful.

sibest:
Wait, I didn't publish everything yet, I will tomorrow (registers/stack capture is missing for non-waiting threads). Anyway, my game seems stable now, finally.

sibest:
Final code is available at https://github.com/sibest/hxcpp, but there are too many changes, will not be committed.

PR #924, opened 2020-07-14
Fixing issue with sockets on windows 

m0rkeulv:
This pull-request will fix an issue with sockets on windows where sockets would not work properly when making debug builds.
The windows compiler for some reason  "fixes" this issue when making release builds  but  the code is still "wrong" in the sense that  logic is executed between the socket error occurs and when the call to get the error code happens.

Debug builds `WSAGetLastError` would always return 0 
https://stackoverflow.com/questions/25438935/winsock-recv-call-returns-1-but-wsagetlasterror-is-set-to-0

### DIFF
--- a/src/hx/libs/std/Socket.cpp
+++ b/src/hx/libs/std/Socket.cpp
@@ -109,11 +109,16 @@ SOCKET val_sock(Dynamic inValue)
 
 static void block_error()
 {
-   hx::ExitGCFreeZone();
+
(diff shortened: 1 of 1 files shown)


m0rkeulv:
@hughsando could you review this one and merge if you approve?

hughsando:
Yes, I guess ExistGcFreeZone may call mutex waits and malloc and quite a few other things. Nice find.

PR #931, opened 2020-09-23
Fix typos

justin-espedal:


Aurel300:
I added these docs to the [manual](https://haxe.org/manual/target-cpp-ThreadsAndStacks.html). I may have fixed some of these typos in the process. Maybe it would be better to PR there?

justin-espedal:
I reviewed the version found in the manual, and all of the typos had already been fixed there. I didn't check any other differences, but I imagine there are other minor structural changes and improvements to the manual version.

Since the documentation is all better maintained in the manual, perhaps it would be better to just copy the changes in the manual over to here whenever they're made. I would suggest just removing these docs and pointing to the manual, but that would risk pointing to documentation for a different version of hxcpp.

hughsando:
It does not make too much sense to correct these in two places.  I did have a concern with keeping the defines in sync with the particular hxcpp version, but it appears there has not been too much action on that front so not really an issue.

PR #932, opened 2020-10-28
Array and map store for 64bit ints

Aidan63:
[haxe pr link]

Added array object for 64bit ints and a 64bit int hash store as well as removing a lot of the Int64 helper stuff.

### DIFF
--- a/include/Array.h
+++ b/include/Array.h
@@ -19,6 +19,7 @@ enum ArrayStore
    arrayFloat,
    arrayString,
    arrayObject,
+   arrayInt64
 };
 
 enum ArrayConvertId
@@ -31,7 +32,8 @@ enum ArrayConvertId
    aciPodBase      = 1,
 };
 
-template<typename T> struct ReturnNull { typedef T type; };
+template<typename T>
+struct ReturnNull { typedef T type; };
 template<> struct ReturnNull<int> { typedef Dynamic type; };
 template<> struct ReturnNull<double> { typedef Dynamic type; };
 template<> struct ReturnNull<float> { typedef Dynamic type; };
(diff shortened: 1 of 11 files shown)


hughsando:
These changes look excellent, but there is a problem because I like to always allow the last (one or two) official releases of haxe to run with bleeding edge hxcpp.  So removing something like the Int64Struct will mean the current haxe compiler will not work.
Generally, adding functions and things is ok - just removing things that is the problem.
There is a solution to this - use the HXCPP_API_LEVEL define.  So in the haxe compiler, update this to generate a new value - generally in line with the next official release, eg "420"
Then in the hxcpp code, eg, `cpp/Int64.h` you can use `#if (HXCPP_API_LEVEL >= 420)` and conditionally use the new code.  If the define is used quite a bit, you could define something like "HXCPP_NATIVE_INT64" in the header and use that instead.
You should then end up with a PR that passes all the test with the old compiler, and when the haxe PR hits, that too will pass.

As I said, these changes look great - I was even thinking of doing this myself soon.  It is ok for Cppia can be inefficient to start with - I will need to add some Int64 stuff to the JIT I think.  If you can't get cppia to work with API 420,  as long as it works with the old compiler, I can merge it in and fix it before I merge the haxe fix in.

Aidan63:
I did wonder about compatibility but wasn't too sure how compatible different haxe and hxcpp versions were. I'll increase the API version and see what parts need to be bought back and wrapped in version check to work with current haxe releases.

Aidan63:
Seems that putting the changed maths functions behind the api level conditional was all that was needed. Tested latest 4.0, 4.1, and the branch with my haxe changes and it seems to work, CI is also passing now as well.

Issue #937, opened 2020-12-17
Can't compile for iOS 


codeservice:
Error message:

Error: /usr/local/lib/haxe/lib/hxcpp/4,0,0/src/hx/Thread.cpp:341:8: error: thread-local storage is not supported for the current target
static thread_local ThreadInfoHolder threadHolder;

hughsando:
Looks like this might be a problem compiling for i386 (simulator) or maybe iOS<version 9.  Is this the case?
The fix is the change the condition `__cplusplus > 199711L` in Thread.cpp to exclude the case where it does not work - just not 100% sure when this is at the moment.

codeservice:
Yes, I was compiling for simulator on latest xCode. Probably possible exclude this case from Thread.

hughsando:
Was it 32 bit simulator? (armv7 for i386, not x86_64 for arm64).  The x86 path is becoming less relevant - not sure it is worth supporting.  Maybe the default simulator is not 64-bit?  That would be sub-optimal.

codeservice:
Compile error rise when I was compiling my project on this step:
haxelib run hxcpp Build.xml -Diphonesim -DHXCPP_M64 -DHXCPP_CPP11 -debug

Compiling group: runtime
xcrun --sdk iphonesimulator14.3 clang++ -D_CRT_SECURE_NO_DEPRECATE -DHX_UNDEFINE_H -c -g -fmessage-length=0 -pipe -Wno-trigraphs -fno-stack-protector -Wno-overflow -fpascal-strings -fasm-blocks -Wreturn-type -Wno-unused-variable -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator14.3.sdk -miphoneos-version-min=5.0 -DIPHONE=IPHONE -DIPHONESIM=IPHONESIM -Wno-parentheses -Wno-null-dereference -Wno-unused-value -Wno-bool-conversion -DHXCPP_DEBUG -DSTATIC_LINK(static) -DHXCPP_M64 -DHXCPP_VISIT_ALLOCS(haxe) -DHXCPP_CHECK_POINTER(haxe) -DHXCPP_API_LEVEL=331(haxe) -DHXCPP_OBJC(haxe) -I/usr/local/lib/haxe/lib/hxcpp/4,0,0/include -fexceptions -fstrict-aliasing ... tags=[haxe]
[...]

PR #938, opened 2020-12-18
Compiling fix for iOS simulator target

codeservice:
Issue description here: https://github.com/HaxeFoundation/hxcpp/issues/937


hughsando:
I think we will go for v9.0 - with a variable to override.  Thanks for the investigation.

PR #942, opened 2021-01-01
Set T* call to 0 when initialized from null

### LINKED ISSUE #941: Setting a Callable to null results in a junk pointer
If callable is set to null, the result is not a null function pointer but a junk pointer:

```haxe
var callable: cpp.Callable<() -> Void> = null;

function main() {
	callable = null;
	trace('callable should be null: $callable');
}
```

traces:
`Main.hx:9: callable should be null: Pointer(0x7ffee8d5a350)`

In `Pointer.h` the wrong assignment operator overload is taken for `callable = null()`, it takes the `Function operator=( const Function &inRHS )` overload

Here are the three assignment overloads
```c++
   inline Dynamic operator=( Dynamic &inValue )
   {;
      call = inValue==null() ? 0 : (T*) inValue->__GetHandle();
      return inValue;
   }
   inline Dynamic operator=( null &inValue ) {
      call=0; return inValue;
   }
   inline Function operator=( const Function &inRHS ) {
      return call = inRHS.call;
   }
```

It does have a `null` assignment overload but it is not taken and I can't figure out why

I'll take a look at resolving this with a PR soon

haxiomic:
Fixes #941

Adds missing `call = 0` when initializing from null (otherwise call is uninitialized and contains junk data)

Before:
```haxe
var callable: Callable<() -> Void> = null;
trace('callable should be null: $callable');
// traces "callable should be null: Pointer(0x10d9f6800)" :[

if (callable != null) callable(); // crash
```

After:
```haxe
var callable: Callable<() -> Void> = null;
trace('callable should be null: $callable');
// traces "callable should be null: null" 👍 
```

This was causing a crash where we were trying to check a Callable was not null before calling

### DIFF
--- a/include/cpp/Pointer.h
+++ b/include/cpp/Pointer.h
@@ -426,7 +426,7 @@ public:
    inline Function( ) { }
    inline Function( const Function &inRHS ) : call(inRHS.call) {  }
    inline Function( const Dynamic &inRHS) { call = inRHS==null()?0: (T*)inRHS->__GetHandle(); }
-   inline Function( const null &inRHS ) { }
+   inline Function( const null &inRHS ) { call = 0; }
    inline Function( T *inValue ) : call((T*)(inValue)) { }
    //inline Function( T *inValue ) : call(inValue) { }
    inline Function( AutoCast inValue ) : call( (T*)inValue.value) { }


haxiomic:
(Test failures unrelated)

hughsando:
Looks good.

haxiomic:
Yey, we're green now 👍

PR #943, opened 2021-02-05
added conditionals for embarcadero c++ (borland)

AdrianV:
In order to compile hxcpp with Embarcadero C++ Compiler, I needed to add these conditionals


hughsando:
Ah,  if defined(__BORLANDC__) brings back some memories!

AdrianV:
Some things will never change :)

PR #946, opened 2021-02-14
GENERATIONAL: Fixes invalid pointers in MarkConservative

sibest:
MarkConservative may cause the GC to fail and crash due to invalid pointers threated as valid pointers inside registers or stack.

The root cause is GetAllocType that is using the non-safe allocStart bitmap to detect allocations. allocStart is only updated in FULL collect passes. A nursery allocation can have the allocStart flag set due to old allocations. In the same time a non-valid allocation can have its allocStart on 1. For this reason GetEnclosingNurseryType must be called without the conditional allocStart check, and
GetAllocTypeChecked must be called after checking if the inOffset is inside an hole.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ hxcpp.n
 *.ilk
 
 .vscode
+*.bak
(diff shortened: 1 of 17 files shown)


Aurel300:
The `.bak` files should not be in the PR.

sibest:
Oops, plus the commit GC_SHORT_ROWS is just for my internal use.

Aurel300:
The `.bak` files are still there; if you want to not commit the `GC_SHORT_ROWS` commit I'd suggest you rewrite history and force push.

PR #956, opened 2021-04-22
Get size of map and regex captures

ALANVF:
Part 2 of an effort to add methods to get the size of a map, and the number of captures from a regex.

There are other ways I could implement the size method for Hashes, so I can do it another way if you don't like my current implementation.

### Option 1
The current solution, which is to cast the `XXXHashBase *` value to `XXXHashObject *` and access `getSize` that way.
```cpp
int __XXX_hash_size(Dynamic &ioHash) {
    XXXHashBase *hash = static_cast<XXXHashBase *>(ioHash.GetPtr());
    if(!hash)
       return 0;
    return ((XXXHashObject *) hash)->getSize();
}
```

### Option 2
Add `HashBase#getSize()` as a virtual method, and remove the `inline` qualifier from the `Hash#getSize()` method.
```cpp
class HashBase ... {
    ...
    virtual int getSize() = 0;
};

class Hash ... {
    ...
    int getSize() { return size; }
};
```

### Option 3
Instead of making `getSize` a virtual method, add a variant of it that *is* virtual instead (that way it doesn't affectexisting code).
```cpp
class HashBase ... {
    ...
    virtual int getDynSize() = 0;
};

class Hash ... {
    ...
    int getDynSize() { return size; }
};
```

### Option 4
Move the `size` property from `Hash` to `HashBase`, and then move `getSize` from `Hash` to `HashBase` so that it can remain inline.
```cpp
class HashBase ... {
    int size;
    ...
    inline int getSize() { return size; }
};

class Hash ... {
    ...
};
```

Part 1: https://github.com/HaxeFoundation/hashlink/pull/437

### DIFF
--- a/include/hx/StdLibs.h
+++ b/include/hx/StdLibs.h
@@ -533,6 +533,7 @@ HXCPP_EXTERN_CLASS_ATTRIBUTES Dynamic _hx_regexp_new_options(String s, String op
 HXCPP_EXTERN_CLASS_ATTRIBUTES bool    _hx_regexp_match(Dynamic handle, String string, int pos, int len);
 HXCPP_EXTERN_CLASS_ATTRIBUTES String  _hx_regexp_matched(Dynamic handle, int pos);
 HXCPP_EXTERN_CLASS_ATTRIBUTES Dynamic _hx_regexp_matched_pos(Dynamic handle, int match);
+HXCPP_EXTERN_CLASS_ATTRIBUTES int     _hx_regexp_matched_num(Dynamic handle);
 
 
 // haxe.zip.(Un)Compress.hx -> src/hx/libs/zlib/ZLib.cpp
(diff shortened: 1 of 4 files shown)


hughsando:
I think option 1/4 can work.  There is already "struct HashRoot" which can form this base class.
The "size" member can be moved into this struct (so could "mask" and "bucketCount")
The dynamic hash could then be static-cast to HashRoot and you would not need a separate function for each type.

ALANVF:
Status?

hughsando:
Looks good, but can you remove the unused type-specific size functions from the header?
Also, the base variables can be exposed in the derived class with `using HashRoot::size` etc in the Hash class, rather than with the `this->` notation.

ALANVF:
Fixed :+1:

hughsando:
Brilliant! Thanks a lot.

PR #962, opened 2021-07-19
Implement non-blocking process exit code fetching

Aidan63:
Adds support for non-blocking process exit code fetching. Should this updated function be behind a `HXCPP_API_LEVEL` ifdef?


Aidan63:
Not sure which pr should be merged first, both depend on a change in the other which is causing the CIs to fail. Guess I'll look into that version check to get the CI green.

Aidan63:
Added an api level check for using the new functions, seems to have got the tests passing.

PR #976, opened 2021-11-25
Semaphores and condition variables.

Apprentice-Alchemist:
See HaxeFoundation/haxe#10503.

### DIFF
--- a/include/hx/Object.h
+++ b/include/hx/Object.h
@@ -122,6 +122,8 @@ enum
    clsIdCStatic5,
    clsIdCStaticVar,
    clsIdMutex,
+   clsIdSemaphore,
+   clsIdCondition,
    clsIdLock,
    clsIdDeque,
    clsIdThreadInfo,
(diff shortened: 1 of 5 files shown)


Apprentice-Alchemist:
Condition variables are not supported on Windows XP and earlier.
Should there be a `HXCPP_WINXP_COMPAT` define that disables condition variables?

Simn:
Could you please push again, I'm not sure what our CI is doing and I can't restart it...

Apprentice-Alchemist:
> I'm not sure what our CI is doing

If you're talking about the linux builds failing, that's because Microsoft removed the ubuntu-16.04 images.
See https://devblogs.microsoft.com/devops/hosted-pipelines-image-deprecation/ and #972.

Simn:
I see, thanks for pointing that out. I have merged #972, so we should be able to green this one now.

Simn:
I'll go ahead and assume that the linux failures are random CI nonsense, let's see if this works after merging...

PR #981, opened 2022-02-12
Fix unicode file operations

tobil4sk:
Certain file system operations were broken when using unicode paths. This fixes them and addresses [haxe#8400](https://github.com/HaxeFoundation/haxe/issues/8400).


Simn:
Thanks!

CI is still stupid but I assume that's unrelated.

PR #985, opened 2022-03-01
Atomic operations.

Apprentice-Alchemist:
See [HaxeFoundation/haxe#10610](https://github.com/HaxeFoundation/haxe/pull/10610)


Simn:
@hughsando Could you take a quick look if this is the right approach for hxcpp?

hughsando:
Do these differ from the versions in include/hx/Thread.h ?

Apprentice-Alchemist:
Oh right, forgot those were a thing.
Let's see, there's more operations, and these only uses intrinsics (those macOS OSAtomic* functions are deprecated).

If you prefer I can merge the two sets of atomic operations.

Simn:
Yes, I think that would be better.

Apprentice-Alchemist:
I'm not sure why the CI hung, that test works fine on my machine....

include/hx/StdLibs.h
```diff
@@ -305,6 +305,114 @@ bool _hx_atomic_exchange_if(::cpp::Pointer<cpp::AtomicInt> inPtr, int test, int
...
+//    Everyone uses GCC, Clang or MSVC
+//    People are not using 8 year old versions of GCC.
+
+#if defined(__GNUC__) || defined(__clang__)
+#define HX_GCC_ATOMICS
+#elif defined(_MSC_VER)
+#define HX_MSVC_ATOMICS
+#include <intrin.h>
+#else // Nearly everyone uses GCC, Clang or MSVC, right?
+#error "Neither GCC, clang or MSVC is being used. Please contribute the relevant atomic instrinsics for your compiler."
```
Simn, line 318: This seems like a good reason to not have this in StdLibs.h.

Apprentice-Alchemist: The GC uses atomics, so the better solution is to fallback to a non-atomic version like the old code.
(The GC code will fall back to a single thread if `HX_HAS_ATOMIC` is 0.)

thomasuster:
Is this what broke android? https://github.com/haxenme/nme/issues/720

PR #988, opened 2022-04-02
Fix hxcpp_parse_int to allow + in hex notation

tobil4sk:
Previously this would result in `null`.

### DIFF
--- a/src/hx/StdLibs.cpp
+++ b/src/hx/StdLibs.cpp
@@ -617,8 +617,8 @@ int __int__(double x)
 
 
 static inline bool is_hex_string(const char *c, int len) {
-   return len >= 2 && c[0] == '0' && (c[1] == 'x' || c[1] == 'X')
-      || len >= 3 && (c[0] == '+' || c[0] == '-') && c[1] == '0' && (c[2] == 'x' || c[2] == 'X');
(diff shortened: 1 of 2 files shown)


Simn:
CI complains about something that seems related:

```
Test Failed:parsed hex value was not 0, null
Test Failed:parsed hex value was not 0, null
```

Not sure what it's even executing there though...

tobil4sk:
Seems to be coming from here:
https://github.com/HaxeFoundation/hxcpp/blob/master/test/std/Test.hx#L915
https://github.com/HaxeFoundation/hxcpp/blob/master/test/std/Test.hx#L935

For:
```haxe
'   0x'
'   -0x'
```
It's about how it handles an empty hexadecimal prefix. I could change this here to return 0 instead of null fairly easily, but in the Haxe unit tests I ended up leaving these cases unspecified because there are some targets where we cannot change this behaviour without overhead.

Simn:
If hxcpp tests this I would prefer to not change the behavior, even if we leave it unspecified in the official docs.

tobil4sk:
Alright, I'll revert it and I'll make sure it hasn't changed on any other targets either.

tobil4sk:
Should be good now. I've added more tests for good measure.

PR #989, opened 2022-04-12
Use absolute path to NDK folders.

player-03:
Since hxcpp doesn't run from the NDK directory, it requires an absolute path to be able to access source.properties. Getting the version from the directory name is unaffected, which is likely why this went unnoticed for so long.

### DIFF
--- a/tools/hxcpp/Setup.hx
+++ b/tools/hxcpp/Setup.hx
@@ -47,7 +47,7 @@ class Setup
           for (file in files)
           {
              file = file.split("\\").join("/");
-             var version = getNdkVersion(file);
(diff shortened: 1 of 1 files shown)


player-03:
Oh yeah, and [this loop](https://github.com/HaxeFoundation/hxcpp/blob/148943c4aad05013894493b6860771842dd091bb/tools/hxcpp/Setup.hx#L331) arbitrarily ends at 19, meaning this error won't happen with newer NDKs. You'll only see the error if testing an old NDK _and_ using the major-minor-micro versioning scheme. (Which is what the Android SDK manager uses.)

PR #1002, opened 2022-07-05
Fix for native extension build errors on MacOS

PaulGene:


Simn:
There's a distinct lack of information here, but adding an additional null-check like this seems sensible in general, so I'll merge this.

PR #1004, opened 2022-07-26
Process: Do not create console on Windows OS

Dimensionscape:
Using the standard api to launch a process on Windows OS creates a console window. With this fix, it should match the behavior of other operating systems.


Dimensionscape:
@hughsando any thoughts on this Hugh?

joshtynjala:
On macOS, no console windows open when using sys.io.Process. It should definitely be the same on Windows. Please merge.

Someone recently mentioned that an app opening up a bunch of console windows looks dangerous to non-developer users (did I get a virus?!). He's not wrong.

Simn:
Makes sense to not have a console here!

PR #1015, opened 2022-11-12
Fix mysql_select_db and mysql_close

AlexHaxe:
after installing a new Linux server I noticed some of my tools failed to connect to a mariadb database.
some debugging and packet sniffing and comparing with PHP communication later I came to the conclusion it's the extra 0x00 byte at the end of database names in `mysql_select_db` that makes it fail.

my changes fix failing mysql tests on my system (while failing on master branch), but I haven't tested Windows or Mac. 
currently they silently fail on azure, so hopefully PR will fix CI as well.

I've also added a `COM_QUIT` packet for `mysql_close` because mariadb always complained about aborted connections since Haxe would just close socket its connection without notifying server side.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -3,6 +3,7 @@
 * Update MIN_IOS_VERSION
 * Bug fixes
 * Upgrade to 4.2 API
+* Fixed mysql_select_db failing for newer mariadb versions due to extra 0x00 byte after database
 
 4.1.15
 ------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Japan50079 (approved):
change internal code

Simn:
Thanks!

Japan50079:
correct

Simon Krajewski ***@***.***>于2022年11月19日 周六18:52写道：

Japan50079:
Just change it

Simon Krajewski ***@***.***>于2022年11月19日 周六18:52写道：

PR #1026, opened 2023-01-20
Update vendored libs

Apprentice-Alchemist:
Vendored libs really should be updated more often, especially for things like mbedtls.

Notes:
1. Has been tested with Haxe test suite, so hopefully that means nothing has broken
2. I haven't updated PCRE since we should update to pcre2, but that's for a future PR.
3. I've updated mbedtls to 2.28.x, but that's only supported until 2024, so at some point we will have to switch to mbedtls 3.x.

Closes #840 (which has been ignored for more than 2 years...)


Uzume:
I am told migration to PCRE2 is not too far in the future, This seems to be the last remaining hold out, see HaxeFoundation/haxe#10491.

PR #1027, opened 2023-02-05
Fix endless recursion in Class_obj::GetInstanceFields

profelis:
cpp::Pointer.mSuper points to same cpp::Pointer instance

Steps to reproduce:
1. get this repo and build debug sample https://github.com/foreignsasquatch/raylib-hx/tree/main/examples
2. add breakpoint inside while loop https://github.com/foreignsasquatch/raylib-hx/blob/main/examples/BasicWindow.hx#L12
3. start debugger session in vscode

stacktrace
https://gist.github.com/profelis/390f80bc47afc16f41b3afced4896cc6

### DIFF
--- a/src/hx/Class.cpp
+++ b/src/hx/Class.cpp
@@ -195,7 +195,7 @@ String Class_obj::__ToString() const { return mName; }
 
 Array<String> Class_obj::GetInstanceFields()
 {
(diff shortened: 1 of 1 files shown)


Simn:
I don't really understand how a pointer relation like this happens in the first place, but this kind of check is usually a good idea anyway, so I'm happy to merge it.

PR #1036, opened 2023-03-22
SSL: fix ssl network error exception when WSAGetLastError() is called too late on Windows

### LINKED ISSUE #1033: "ssl network error" thrown during non-blocking sys.ssl.Socket handshake when compiled with -debug and -D HXCPP_M64 on Windows
### Overview:

Example code:

```haxe
import haxe.Timer;
import haxe.io.Error;

class Main {
	static function main() {
		var socket = new sys.ssl.Socket();
		try {
			socket.setBlocking(false);
			socket.connect(new sys.net.Host("haxe.org"), 443);
		}
		catch (e:Dynamic) {}
		function tryHandshake():Void {
			var result = sys.net.Socket.select(null, [socket], null, 0);
			if (result.write[0] != socket) {
				Timer.delay(tryHandshake, 16);
				return;
			}
			try {
				socket.handshake();
			}
			catch (e:Error)
			{
				switch (e)
				{
					case Error.Blocked:
						Timer.delay(tryHandshake, 16);
					case Error.Custom(Error.Blocked):
						Timer.delay(tryHandshake, 16);
					default:
						trace("handshake failed 1: " + e);
				}
				return;
			}
			catch (e:Dynamic)
			{
				trace("handshake failed 2: " + e);
				return;
			}
			trace("handshake succeeded");
		}
		Timer.delay(tryHandshake, 16);
	}
}
```

Compile command:

```sh
haxe -cp src -main Main -debug -cpp out -D HXCPP_M64
```

If you remove one of `-debug` or `-D HXCPP_M64` (or both), the handshake completes successfully. If the socket is blocking, the handshake completes successfully.

I confirmed that it works correctly with all combinations on macOS.

### System info:

Haxe 4.2.5
hxcpp 4.2.1
Windows 11 Pro 22H2

joshtynjala:
Fixes #1033

### DIFF
--- a/src/hx/libs/ssl/SSL.cpp
+++ b/src/hx/libs/ssl/SSL.cpp
@@ -181,16 +181,19 @@ struct sslpkey : public hx::Object
 static mbedtls_entropy_context entropy;
 static mbedtls_ctr_drbg_context ctr_drbg;
 
+static bool is_ssl_blocking( int r ) {
+	return r == MBEDTLS_ERR_SSL_WANT_READ || r == MBEDTLS_ERR_SSL_WANT_WRITE;
+}
 
-static void block_error() {
+static bool is_block_error() {
 	#ifdef NEKO_WINDOWS
 	int err = WSAGetLastError();
 	if( err == WSAEWOULDBLOCK || err == WSAEALREADY || err == WSAETIMEDOUT )
 	#else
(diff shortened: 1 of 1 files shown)


joshtynjala:
Thank you, @hughsando!

hughsando:
Thank You!

PR #1040, opened 2023-04-08
Replace `<::cpp::Int64>` with `< ::cpp::Int64>` to fix template errors

jeremyfa:
On hxcpp 4.3.2, I got template build errors related to usage `<::cpp::Int64>` when using native code that includes <hxcpp.h> like the one in [linc_dialogs](https://github.com/ceramic-engine/linc_dialogs).

It seems the way to go is to simply keep a space between `<` and `::` so that the compiler doesn't parse it wrong. This pull request does just that.

@Aidan63 you might be interested in this PR as it's related to your Int64 changes.

### DIFF
--- a/include/Array.h
+++ b/include/Array.h
@@ -44,7 +44,7 @@ template<> struct ReturnNull<unsigned char> { typedef Dynamic type; };
 template<> struct ReturnNull<short> { typedef Dynamic type; };
 template<> struct ReturnNull<unsigned short> { typedef Dynamic type; };
 template<> struct ReturnNull<unsigned int> { typedef Dynamic type; };
-template<> struct ReturnNull<::cpp::Int64> { typedef Dynamic type; };
+template<> struct ReturnNull< ::cpp::Int64> { typedef Dynamic type; };
 
 template<typename T>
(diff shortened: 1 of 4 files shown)


Aidan63:
Good spot, seems like this error is related to the C++ standard version. Prior to C++ 11 `<::` would be treated as a digraph, but C++ 11 added an extra rule stating that it should not be treated as a digraph if the next character is not `:` or `>`.

> Otherwise, if the next three characters are <:: and the subsequent character is neither : nor >, the < is treated as a preprocessor token by itself and not as the first character of the alternative token <:.

Not sure if hxcpp has a miniumum C++ version it wants to supports, but adding those spaces are an easy fix to get it working pre C++ 11.

jeremyfa:
Yes, I have seen other places where a space is used, like with: `< ::String>`, so I came up with the same workaround here

Aidan63:
I remember seeing those spaces before `<` and always wondered if that was a style choice or actually required, since it seemed to work fine for me I assumed it was the former, but now I know!

I wonder if my changes to generic extern classes could generate this sort of invalid C++...

PR #1042, opened 2023-04-11
Add -std=c99 when compiling pcre2 sources

### LINKED ISSUE #1041: PCRE2 requires C99 support
This is from Lime's github actions android builds:

```
hxcpp/4,3,2/project/thirdparty/pcre2-10.42/src/pcre2_script_run.c:109:1: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
 for (int i = 0; i < FULL_MAPSIZE; i++) require_map[i] = 0;
 ^
note: use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code
```

Seems like the new version of pcre from #1037 doesn't compile with older c standards.

It looks like this is why this commit was necessary? 27210f2160c1138cfde433ff2f384383928afa0d @hughsando 

Maybe we can set the flag only for the pcre sources? Otherwise, it requires adding the c99/c11 flag to the remaining toolchains.

tobil4sk:
Closes #1041


hughsando:
Thanks!

PR #1043, opened 2023-04-11
Add tools/hxcpp/cpp to git ignore

tobil4sk:


tobil4sk:
@Simn Green here now as well (although this was pretty trivial anyway).

PR #1047, opened 2023-04-15
Draft: Callable object for typed functions

Aidan63:
See https://github.com/HaxeFoundation/haxe/pull/11151 for details

### DIFF
--- a/include/hx/Functions.h
+++ b/include/hx/Functions.h
@@ -23,6 +23,178 @@ namespace hx
 #endif
    };
 
+    template<class TReturn, class... TArgs>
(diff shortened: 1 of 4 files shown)


Aidan63:
These failing GC tests are very odd. Take the `testFunc` test, the object passed into the do not kill is correctly identified as a closure object, however after a GC collection retreiving that object via the zombie list gives an anon object when inspecting it in the visual studio debugger?? There are no other objects in the list and oddly this only occurs with my hxcpp changes on 4.3.1...

This also only occurs on x64 windows with the debug flag, remove the debug flag and all four failing GC tests pass.

PR #1053, opened 2023-04-26
`__hxcpp_parse_int` now parses hex numbers with wrapping as before

### LINKED ISSUE #1052: `__hxcpp_parse_int` parsing Hexadecimal string returns the wrong value
I recently noticed this when I was working on a new feature for a mod I'm working on. Whenever entering any hex value, i.e `0xFF000000`, or `0xFF00FFFF`, or any hex value like that, it gives the same translucent white color, with a value of 2147483647 which is the limit for signed 32 bit integers.

I'm unsure why this happens; but my first assumptions are that it roots from commit d204bca99ddd51ba3ebcaa35643054512b31defa as I do not remember having these issues with the function before this commit, as replacing StdLibs.cpp with the file from b8248a719f45a1c3989f3d6c998c863724d18925 seems to have no issues.

To further reinforce the point above, using StdLibs.cpp from b8248a719f45a1c3989f3d6c998c863724d18925 returned the correct values for both colors I stated above as examples.

Here's what happens when trying to parse "0xFF000000" (`var color:String = "0xFF000000"`) and "0xFF00FFFF", with the int value itself and the parsed string:

image

Via commit b8248a719f45a1c3989f3d6c998c863724d18925:
0xFF000000
image
0xFF00FFFF
image

Via commit d204bca99ddd51ba3ebcaa35643054512b31defa:
0xFF000000
image
0xFF00FFFF
image

This shows that commit b8248a719f45a1c3989f3d6c998c863724d18925 prints the correct value but commits d204bca99ddd51ba3ebcaa35643054512b31defa and higher return 2147483647. I'm not the most well-versed in C++, so I cannot figure out why this happens myself, but I hope this is enough information to figure out and fix this issue.

Any help would be greatly appreciated.

SomeoneSom:
`__hxcpp_parse_int` now parses hex numbers with wrapping as before, fixing a regression.

Old:
```haxe
Std.parseInt("0xFFFFFFFF"); // 2147483647
```

New:
```haxe
Std.parseInt("0xFFFFFFFF"); // -1
```

Closes #1052

### DIFF
--- a/src/hx/StdLibs.cpp
+++ b/src/hx/StdLibs.cpp
@@ -633,7 +633,16 @@ Dynamic __hxcpp_parse_int(const String &inString)
    while (isspace(*str)) ++str;
    bool isHex = is_hex_string(str, strlen(str));
    char *end = 0;
-   long result = isHex ? strtoul(str,&end,16) : strtol(str,&end,10);
+   long result;
+   if (isHex)
+   {
+      bool neg = str[0] == '-';
+      if (neg) str++;
+      result = strtoul(str,&end,16);
+      if (neg) result = -result;
+   }
+   else 
+      result = strtol(str,&end,10);
    #ifdef HX_WINDOWS
    if (str==end && !isHex)
    #else
--- a/src/hx/StdLibs.cpp
+++ b/src/hx/StdLibs.cpp
@@ -633,7 +633,7 @@ Dynamic __hxcpp_parse_int(const String &inString)
    while (isspace(*str)) ++str;
    bool isHex = is_hex_string(str, strlen(str));
    char *end = 0;
-   long result = strtol(str,&end,isHex ? 16 : 10);
+   long result = isHex ? strtoul(str,&end,16) : strtol(str,&end,10);
    #ifdef HX_WINDOWS
    if (str==end && !isHex)
    #else


tobil4sk:
Hm... this looks to me like it would break negative hex parsing `Std.parseInt("-0x1");` ? But oddly, according to the tests it doesn't...

SomeoneSom:
The truncation to int makes the overflow happen correctly, since the upper bits should just get discarded if they exist (though this does seem to be implementation undefined).
```haxe
0xFFFFFFFFFFFFFFFF // result of strtoul("-0x1", NULL, 0) where an unsigned long is 64 bits, 18446744073709551615
0x--------FFFFFFFF // result of truncating that to an int, where an int  is 32 bits, -1
```
The fact that the cast is implementation specific though could have unforeseen consequences though, there might be some safer way to cast while forcing the truncation.

SomeoneSom:
Looking at the previous iteration that used `strtoul`, it manually negated the result. I'll just implement that.

SomeoneSom:
Manual negative hex parsing has been added. I ran the integer parsing tests on my machine locally, so the tests should hopefully pass.

tobil4sk:
Thanks! Looks good now.

PR #1078, opened 2023-12-20
Use hx4compat for running sys.db tests

tobil4sk:
Fixes errors in test suite:
```
Test.hx:5: characters 8-25 : Type not found : sys.db.Connection
Test.hx:6: characters 8-21 : Type not found : sys.db.Sqlite
Test.hx:7: characters 8-20 : Type not found : sys.db.Mysql
Test.hx:102: characters 38-48 : Type not found : Connection
```

### DIFF
--- a/test/std/compile32.hxml
+++ b/test/std/compile32.hxml
@@ -3,3 +3,4 @@
 -D HXCPP_M32
 -D HXCPP_DEBUGGER
 -cp ../unit
+-lib hx4compat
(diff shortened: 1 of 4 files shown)


Simn:
Looks like it's still missing somewhere.

tobil4sk:
That job only compiles `test/std/compile*.hxml` which definitely all include hx4compat... Perhaps there is something wrong with the class resolution now that the classes are no longer in `std/cpp/_std`? Or maybe this error comes from the macro context or something?

There is no line number given with the error, unlike before:
`Type not found : sys.db.Sqlite`

Simn:
No idea what the mac tests are doing, but this is certainly better than before. Thanks!

PR #1083, opened 2023-12-31
Remove Haxe 3 Code

### LINKED ISSUE #1081: HXCPP_API_LEVEL 300 Cleanup
It would be nice if we could remove all the haxe 3 paths in the code base, some areas have quite the `#ifdef` collection which can make navigating and understanding whats going on harder than it needs to be.
I remember you said you like to keep support for the last two releases but I'm not sure if thats major versions (e.g. haxe 3 and 4) or sub versions (e.g. haxe 4.3 and 4.2). Either way with haxe 5 seemingly coming next it would be good to remove this old code.

Thanks.

Aidan63:
Closes #1081


jdonaldson:
Nice work cleaning this all up!

Issue #1082, opened 2023-12-29
GC Pinning


Aidan63:
We can root objects, but from what I can tell we can't pin to prevent them from being moved by the GC between collections.
My use case for this is wanting to make it easier to integrate with async libraries like libuv, currently if I want to write haxe bytes to a file through libuv I need to copy the data into a non GC staging buffer and write that otherwise the GC could move the haxe object mid write.
I've not really dug around the GC before but I can see a commented out `IMMIX_ALLOC_IS_PINNED` define which is not referenced anywhere else, was this functionality planned or existed in a previous version?

Thanks.

hughsando:
Currently, the only pinning that happens is when a pointer is found on the stack.  In this case, the whole allocation block that contains the pointer is pinned. 
Note that if you store stuff in a GC root, then the pointer will be updated - however, this update will happen synchronously for haxe threads, but it will be async for foreign threads so perhaps not suitable.
Note also that most of the APIs are for hx::Object* but in the case of a byte array, it will be the bytes (contents) that need to be pinned, not the Array<Byte> object (this will need to be rooted to prevent collection rather than moving). The  byte itself is not an instance of  hx::Object.

If the byte array is big enough (IMMIX_LARGE_OBJ_SIZE = 4000), it will be allocated separately from the block system and will not need to be pinned.  This perhaps offers a reasonable solution - copy the data if is is smaller than this value or just use the pointer if it is bigger.

If you don't like this, then code will need to be added - either a list of pinned pointers, which get run though after "MarkAll" in the moving case to pin the blocks, or a "pinCount" in the BlockDataInfo structure and can be increased/decreased externally and can be used to initialize the "mPinned" variable to true in clearBlockMarks.

Aidan63:
Hi Hugh,

Thank for the info. What I've been doing for large byte arrays is to root it and then copy chunks of the arrays data into a `std::vector` and give the data pointer of that to libuv to write. In the libuv callback I then copy the next chunk from the original haxe array into that vector and send to write again and once there is nothing else to write the array is un-rooted.
This works fine but I need to then keep a queue for writes and flushes so everything is done in order, which adds a fair bit of code complexity.

[...]

PR #1085, opened 2024-01-03
Expose MemType

Aidan63:
Makes the `MemType` enum and `GetMemType` function accessible from the `hx` namespace as mentioned in #1082.

### DIFF
--- a/include/hx/GC.h
+++ b/include/hx/GC.h
@@ -283,6 +283,10 @@ void GCPrepareMultiThreaded();
 namespace hx
 {
 
+enum MemType { memUnmanaged, memBlock, memLarge };
+
+MemType GetMemType(void*);
+
 #define HX_USE_INLINE_IMMIX_OPERATOR_NEW
 
 //#define HX_STACK_CTX ::hx::ImmixAllocator *_hx_stack_ctx =  hx::gMultiThreadMode ? hx::tlsImmixAllocator : hx::gMainThreadAlloc;
--- a/src/hx/gc/Immix.cpp
+++ b/src/hx/gc/Immix.cpp
@@ -2907,8 +2907,6 @@ typedef hx::QuickVec<BlockDataInfo *> BlockList;
 
 typedef hx::QuickVec<unsigned int *> LargeList;
 
-enum MemType { memUnmanaged, memBlock, memLarge };
-
 
 
 
@@ -5524,7 +5522,7 @@ public:
       return false;
    }
 
-   MemType GetMemType(void *inPtr)
+   hx::MemType GetMemType(void *inPtr)
    {
       BlockData *block = (BlockData *)( ((size_t)inPtr) & IMMIX_BLOCK_BASE_MASK);
 
@@ -5542,16 +5540,16 @@ public:
       */
 
       if (isBlock)
-         return memBlock;
+         return hx::memBlock;
 
       for(int i=0;i<mLargeList.size();i++)
       {
          unsigned int *blob = mLargeList[i] + 2;
          if (blob==inPtr)
-            return memLarge;
+            return hx::memLarge;
       }
 
-      return memUnmanaged;
+      return hx::memUnmanaged;
    }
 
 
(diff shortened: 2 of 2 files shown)


hughsando:
I think there are some complications with using the GC memtype routines outside of a collection (ok in __Mark).
For example, the large list might change - it is locked inside a collect call, but not in general.
On the plus side, if you know it is a valid pointer inside a haxe object, then you can rely on the header bits, and it should not be memUnManaged except in case of Array.mAlloc=-1 .

With array data, it is possible that the the pointer is a foreign pointer (Set with NativeArray.setUnmanagedData).  In this case, the arrays mAlloc will be -1, and no memory management on the pointer is needed - this should be checked before examining the header.

You can check for 'const-ness' with bool IsConstAlloc(const void *inData).  In this case, the data does not need any management at all (eg, string data) and will not be deleted.

largeness can be determined from a valid header from the size ("size in allocated blocks"), which will be zero in this case
```cpp
 int size = flags & 0xffff;
 // Size will be 0 for large allocs -> no need to mark block
```

You may need to add some code to prevent large allocations from getting destroyed in a realloc process, see InternalRealloc and InternalReleaseMem.  The theory is that only 1 array will hold the byte data, so it can be cleaned when the array is resized.
This may need some thought, but one possibility could be set the least-significant for the large array size to 1 in the case where we do  not want immediate reallocation.

If you are storing GC block pointers in a haxe object, some additional work must be done the in the generational case, HX_OBJ_WB should be called.  This is for the case where the object holding the pointer is "old" and the pointer is generationally new.  This call means your __Mark will get called when doing a minor generational collection.

Aidan63:
Thanks for the tips, I see what you mean about the large list locking.

Perhaps it makes sense to come at this from a different angle then, if arrays had a `pin` method which returned some sort of handle it could keep track if a pin was requested and not attempt that realloc optimisation if it was. That pin method could do all the checking to see if `mBase` is unmanaged, lives on the LOH, or in a block and act accordingly. The handle could keep a rooted marking object which calls `hx::MarkAlloc` in its `__Mark` to keep that object alive until the handle is released.

Aidan63:
OK, I've completely re-jigged this merge. MemType changes have been reverted and it adds pinning spport.
Pinning and unpinning checks the header to see if it lives on the LOH and `BlockDataInfo` now keeps track of the number of allocs in its lines and sets `mPinned` accordingly.
Array now has a pin function which will return a handle keeping the array storage at time of pinning alive and un-moving until the handle is deleted.
Hopefully this is less dodgy.

hughsando:
Probably better to use the atomic inc/dec ops for mPinned manipulation if pinning might happen from any haxe thread.
It would be better if possible to not increase the size of the Array class since this will be allocated a lot, and pinning will happen much less often.  I was thinking a std::unordered map of pinned pointers might do the trick.  These could be marked like the conservative marks - pinning the blocks at mark time.  If you reuse the mLargeListLock for the pinned set you could solve the realloc problem by checking the pinned set inside FreeLarge for almost no cost.  Marking the list like conservative pointers also solve the generational gc problem.
You still might need your pinned structure to allow unpinning of exactly the same pointer that was pinned (since the array might realloc in the meantime).  You could possibly just use a cpp.Star pointer in a normal haxe structure to do this though.

Aidan63:
I thought of the atomic operations just as I was going to bed, I've changed things around now though so that `mPinnedAllocs` isn't needed any more.
There is now an `sgPinSet` of `void*` pointers which is protected by the `mLargeListLock` of the global allocator. Just before the roots are traversed this set is iterated over and any non object not on the LOH has its block pinned. `FreeLarge` now checks the pin set after aquiring `mLargeListLock` and exits early if the pointer to be released is in the set.
One thing I have noticed is the shrinking case of arrays, it doesn't re-allocate but just updates the length value. However, it also memsets the now extra bytes to 0 which is a problem when wanting to pass that data around.

hughsando:
This seems good.  Changing the length is the same as changing, say, the first element - more of a documentation issue if it does not crash. The only alternative is to copy, since there is not real way to do copy-on-write without incurring overhead.
The onus is on the user to ensure that the unmanaged pointers remain valid, so no  problems there.

Aidan63:
Yeah, thats fair. The only way to solve it without the extra copy would probably be to introduce an extra BytesData type instead of that being a typedef to an array. Since this is primarially going to be used with haxe bytes that would stop the user peeking at the underlying implementation and resizing it.
But thats a lot of effort (and a breaking change) for something which the user should know is asking for trouble (haxe bytes don't expose any way in the formal API for resizing).

Aidan63:
Probably should have opened this as a draft to begin with. I'm not planning on adding anything else so if this looks good its probably good to merge.

PR #1088, opened 2024-01-26
Fix arm64 library names on mac

tobil4sk:


tobil4sk:
@Simn green here now as well! Thanks.

PR #1091, opened 2024-02-19
Fix android for large builds

thomasuster:
TLDR, this fixes large android builds from crashing during the link step.

Details:
It took me a very long time to get this working, and I wasn't able to discover why exactly this occurs. What is interesting about this step of the build is that the output for the program I'm linking is large, >200,000 characters. So originally I tried making the hxcache dir path as short as possible. When that didn't work I looked into ulimits, launchctrl, and tried looking at any kind of diagnostics that OSX has builtin. Ultimately I was unsuccessful. As a shot in the dark, I tried writing the command HXCPP was about to run to a temporary .sh file, and then invoked that sh file and was surprised to see that it worked. To me that signaled perhaps the problem is the stdout / stdin trying to get the data back into the NekoVM. There is probably some more beautiful way to fix this, either in neko VM, or some bizaare OSX environment option I wasn't able to find.


hughsando:
Thanks!  This fix has been used.  Android/clang does some funny stuff with scripts, so "clang++" is actually a script that runs the real exe, so something might be going wrong with this.  But the -v is not really needed, so good fix.

PR #1095, opened 2024-03-07
Make sys_read_dir errors consistent between platforms

tobil4sk:
Haxe documentation specifies that an exception should be thrown if the path is not a valid directory:

> If `path` does not denote a valid directory, an exception is thrown.

This was already the case on all platforms except Windows, where `null` was returned. Now the behavior is consistent.


tobil4sk:
@Simn This one also seems to be affected by the flaky test.

tobil4sk:
Looks green now!

PR #1104, opened 2024-04-25
Fixed String.fromCharCode when using reflection

### LINKED ISSUE #1103: Reflect.field(String, "fromCharCode") doesnt work
https://github.com/HaxeFoundation/hxcpp/blob/master/src/String.cpp#L2176
Hxcpp implementation of String is missing this

```cpp
bool String::__GetStatic(const ::String &inName, Dynamic &outValue, ::hx::PropertyAccess inCallProp)
{
	if (HX_FIELD_EQ(inName,"fromCharCode") ) { outValue = fromCharCode_dyn(); return true; }
	return false;
}
```

which causes String.fromCharCode to not work on reflection.

NeeEoo:
Fixes #1103 

Sorry if the code is bad, i don't know the proper way to do this. Feel free to edit this pr.

```hx
try {
    Reflect.callMethod(null, Reflect.field(String, "fromCharCode"), [65]);
    Sys.println("String.fromCharCode works from reflect.");
} catch(e:Dynamic) {
    Sys.println("[ERROR] String.fromCharCode doesn't work from reflect.");
}
```

### DIFF
--- a/include/hxString.h
+++ b/include/hxString.h
@@ -354,6 +354,9 @@ public:
    // This is used by the string-wrapped-as-dynamic class
    hx::Val __Field(const ::String &inString, hx::PropertyAccess inCallProp);
 
+   // Allows for reflection to be able to get the static functions
+   static bool __GetStatic(const String&, Dynamic&, hx::PropertyAccess);
+
    // The actual implementation.
    // Note that "__s" is const - if you want to change it, you should create a new string.
    //  this allows for multiple strings to point to the same data.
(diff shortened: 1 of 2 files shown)


MAJigsaw77:
Any progress on this?

NeeEoo:
Is there any problems with this pr, that has caused it to not be merged yet?

Raltyro:
I want to reply to this pull request that the author are no longer interacting with the internet, so this is a dead pull request.

tobil4sk:
@Aidan63 I think we could merge this, before conflicts arise. The haxe PR will add a test but it would have to be fixed for other targets first before that can be merged.

MRK2009:
> I want to reply to this pull request that the author are no longer interacting with the internet, so this is a dead pull request.

Because some drama happened to him Ig (Which I don't want to talk about here).

PR #1106, opened 2024-04-29
GitHub CI

Aidan63:
Been wanting this for a while since forks don't inherit azure pipeline stuff, so here's the existing azure pipeline CI ported to github actions. Main changes are as follows;

- Cppia tests were not ran on the CI previously, they are now.
- The individual test suites are now ran as independent jobs meaning they're done in parallel, this gives the CI a nice speed up. Another benefit is seen with that flakey mac hostname test, now it will report std tests are failing, but still run all the others.
- I restricted the recent github CI additions (artefacts and tagging) to only run on the master branch. Previously it ran on every pushed commit of every branch, which I'm guessing wasn't the intention with this?

Example of what a branch of mine looks like, all those tags were generated by the CI.
image

Some small things I've noticed, the "haxe" tests for 32bit windows take an age to run (15mins vs 5mins for 64bit Windows and other OS'). Building the haxe repo test suite also takes 16mins on Mac vs ~4mins on Windows and Linux.


Aidan63:
I just forgot to add a pull request trigger so nothing ran when I opened it. It seems to have kicked into action now.

Aidan63:
Seems like the checks tab just shows a big list of the 100 spawned jobs, shame it doesn't show something more structured like it does in the repo actions UI.

image

Aidan63:
I've split the existing github actions script for packaging back into its own file. So thats now the same as the one in master with the only change being it now only runs on pushes to master, which is probably along the lines of the original intention of it.

One new development is that haxe nighlies no longer seem to work on osx? The setup haxe action always fails when it tries to setup haxelib, it complains about not being able to find libneko.

```
 /usr/bin/tar xz -C neko-2.3.0-osx64 -f /Users/runner/work/_temp/d41fcca9-6a05-40e9-968e-3aeac137eed5
/bin/ls -1 neko-2.3.0-osx64
neko-2.3.0-osx64
/usr/bin/tar xz -C haxe_latest -f /Users/runner/work/_temp/d00a9bdc-8d92-469c-8e25-93d463847701
/bin/ls -1 haxe_latest
haxe_20240612100642_2106be4
/bin/ln -sfv /Users/runner/hostedtoolcache/neko/2.3.0/x64/libneko.2.dylib /Users/runner/hostedtoolcache/haxe/latest/x64/libneko.2.dylib
/Users/runner/hostedtoolcache/haxe/latest/x64/libneko.2.dylib -> /Users/runner/hostedtoolcache/neko/2.3.0/x64/libneko.2.dylib
/Users/runner/hostedtoolcache/haxe/latest/x64/haxelib setup /Users/runner/hostedtoolcache/haxe/latest/x64/lib
dyld[3399]: Library not loaded: @rpath/libneko.2.dylib
  Referenced from: <7B493A09-7C35-3D4A-8AE7-A0A7B110E9EB> /Users/runner/hostedtoolcache/haxe/latest/x64/haxelib
  Reason: tried: '/usr/local/lib/libneko.2.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/local/lib/libneko.2.dylib' (no such file), '/usr/local/lib/libneko.2.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/local/lib/libneko.2.dylib' (no such file)
```

@tobil4sk you've been doing some haxelib and neko packaging stuff, right? Anything changed fairly recently which could cause this? Not sure if this is a problem with the haxe setup action, haxelib, neko, or something else...

tobil4sk:
> Anything changed fairly recently which could cause this?

I think this is due to https://github.com/HaxeFoundation/haxe/pull/11653. Previously the haxelib binary had the same rpath as the neko binary, which included @loader_path. However, now the haxelib binary is built with its own rpath which does not include this. This is why it is failing to find libneko. I'm not sure if the binaries are intended to be installed the way that setup-haxe is doing it.

This can be fixed either by adding the rpath to the haxelib binary or by changing setup-haxe to install the binaries to a default loader path (`/usr/local/lib`) or to modify the binary with `install_name_tool` so it can locate libneko in the custom install path.

Aidan63:
Thanks, I'll have a look at haxe setup and see if I can update it.

Aidan63:
Getting neko to work again was as simple as ensuring `DYLD_FALLBACK_LIBRARY_PATH` points to where neko is, I've also updated setup-haxe to work on the latest ARM mac runners, so this repo and any others don't need to use macos-13, the last runner to use x86 hardware. I've updated this branch to point to my setup-haxe fork for now.

https://github.com/krdlab/setup-haxe/pull/47

https://github.com/krdlab/setup-haxe/pull/48

I think this is good to go now, but we might want to wait for those to get merged and have a new setup-haxe release instead of it pointing to a commit in my fork.

Simn:
I feel like there has been enough waiting, we can always update this later if those PRs get merged. Thank you!

PR #1107, opened 2024-05-10
Use cwd when running `haxelib path`

tobil4sk:
If a local haxelib repo is being used, then moving out of the cwd will break library resolution.

This was presumably done in case `haxelib` is not in `PATH`, but in that situation `HAXEPATH` is unlikely to have been set either.

Fixes: https://github.com/HaxeFoundation/haxelib/issues/618

### DIFF
--- a/tools/hxcpp/PathManager.hx
+++ b/tools/hxcpp/PathManager.hx
@@ -106,7 +106,7 @@ class PathManager
          
          try
          {
-            output = ProcessManager.runProcess(Sys.getEnv ("HAXEPATH"), "haxelib", [ "path", name ], true, false);
+            output = ProcessManager.runProcess("", "haxelib", [ "path", name ], true, false);
          }
          catch (e:Dynamic) {}
          


Simn:
Shouldn't this use `Sys.getEnv ("HAXEPATH") ?? ""` instead? It might be unlikely that `HAXEPATH` is set, but it's still possible.

tobil4sk:
>  It might be unlikely that `HAXEPATH` is set, but it's still possible.

The problem is that we don't want it to change directories, otherwise if there is a local repo in the project directory it will change directory and so the local repo won't get used, like in https://github.com/HaxeFoundation/haxelib/issues/618.

I think originally it changed directory to `HAXEPATH` in case `haxelib` isn't found in `PATH`. But what I was saying is that if `haxelib` is not found in `PATH`, then something probably went wrong in the installation process, so in that case `HAXEPATH` probably won't have been set anyway. So changing CWD to `HAXEPATH` doesn't seem to solve the problem it was trying to address, so it seem safe to change it to `""` which will just keep it in the project directory. This way it respects the project's local repo.

Simn:
Thanks for the explanation, makes sense!

PR #1115, opened 2024-06-16
CI - Host and Localhost tests

Aidan63:
The Mac CI has been failing for quite some time now due to a host address resolution test which tries to resolve the localhost. Resolving localhost addresses on mac runners on github actions and azure pipelines has been broken for the last year or so.

https://github.com/actions/runner-images/issues/8649

To work around this I've split up the tests. Using `Host.localhost` for testing if we can get the localhost address, and `new Host("github.com")` to test if we can resolve an address. Github should be a pretty safe url to test against.

With this hopefully the CI will be consistently green again!

### DIFF
--- a/test/std/Test.hx
+++ b/test/std/Test.hx
@@ -420,17 +420,30 @@ class Test
 
    }
 
-   public static function testHost()
+   public static function testLocalhost()
    {
       log("Test Host");
       try
       {
       var localhost = Host.localhost();
       v('localhost :$localhost');
-      var host = new Host(localhost);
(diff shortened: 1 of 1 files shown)


tobil4sk:
@Aidan63 Note, `testLocalHost()` doesn't get run because it has to be called explicitly in the `main()` method in `test/std/Test.hx`. Was it omitted intentionally?

Aidan63:
Ah, no, that was an oversight on my part then. Must have assumed it was a utest project so any function starting with `test` would be auto ran. I'll get a fix for it this evening.

PR #1118, opened 2024-06-25
Add HXCPP_CPP17 to mingw, ios and android toolchains

tobil4sk:
Follow up to #1116


Simn:
Feel free to ping me when any of those PRs are merge-ready! It's sometimes not easy to tell and I don't want to merge something too early again.

I think #1106 requires a rebase for all PRs anyway to be safe.

tobil4sk:
Looks like there is a failing windows test here, but it looks like a flaky test so maybe rerunning will fix it:

https://github.com/HaxeFoundation/hxcpp/actions/runs/9662043222/job/26651134945?pr=1118

PR #1119, opened 2024-06-25
CI - Stop master releases using existing version numbers

Aidan63:
The counter the CI release logic uses was reset when I moved the yml to a different workflow file so the last two master commits have overwritten previous github releases... I've worked around it in this by just adding a number larger than the latest pre-github CI merge.
As mentioned in the block comment, this will need to be looked at again when we want anything other than the last build number to increase.

@Simn

### DIFF
--- a/tools/version/Write.hx
+++ b/tools/version/Write.hx
@@ -5,7 +5,14 @@ class Write
    public static function main()
    {
       var args = Sys.args();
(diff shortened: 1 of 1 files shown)


Aidan63:
Update: Seems like it will fail instead of overwrite when it uses the same version number, which is good.

Issue #1060, opened 2023-07-01
GC generational zombie tests failing with x64 MSVC and C++17 in debug mode


Aidan63:
Originally spotted when the CI started failing in my #1047 request, but when running the hxcpp tests with haxe 4.3.1 on MSVC for HXCPP_M64 with the debug flag set and the msvc toolchain modified to enable C++17 `<flag value="/std:c++17"/>`, the `testObject`, `testFunc`, `testCustomObject`, and `testBytes` zombie tests start to fail. The last three of these seem to be a knock on from the first failing test.

`testFunc` is failing with the zombie from the GC being null. Stepping through things with visual studio, when `FindZombies` is called during collection the object marked as do not kill is never moved into the zombie list as it's mark byte has been set.
Assuming that this is due to some new aggressive optimisation MSVC can now do I started to modify the code a bit, seeing that `clearStack` has a `nothing` argument which, as the name suggests, does nothing I added a few more assuming that it exists to avoid compiler optimisations. Adding four more nothing arguments to that `clearStack` function then gets all four of those tests to start passing again! It is odd that this only happens with the debug flag, I'd have though thses sort of optimisations would be reserved for release modes.

I'm planning on not requiring C++17 for #1047, but its something to be aware of atleast...

PR #1120, opened 2024-06-25
CI - UTest2

Aidan63:
@tobil4sk noticed that the change I made to avoid the flakey Mac test wasn't actually running. I assumed the std tests were using utest, but no, the results are manually collected.
Initially I was just going to get that new test to run, but then I thought "in for a penny in for a pound" and decided to move those std tests to utest. Then I discovered hxcpp doesn't use utest and instead has a vendored version of the old haxe 3 test stuff.

So here's the hxcpp test suite migrated to UTest2 / utest git.


Aidan63:
Seems like these are legitimate failures and I can recreate them, will investigate.

Aidan63:
Fixed the few failing tests, both seem to be variants of what seems to be new optimisations done my MSVC triggered by the test suite code changes. I've previously seen this with the GC tests which fail when compiling with C++17 on Windows (https://github.com/HaxeFoundation/hxcpp/issues/1060).

I've worked around this by adding more dummy `?_:Dynamic` arguments which seems to cancel out whatever stuff MSVC is doing and gets those tests to pass as they were before.

Issue #1121, opened 2024-06-25
Mac CI takes an age to run


Aidan63:
Compiling the haxe test suite on macos runner is very slow compared to Windows and Linux.

- Windows: around 2mins 30secs
- Linux: around 5mins
- Mac: half an hour!

This seems to be a github issue (which they are very quiet about) so I don't know if there's anything that can be done.

https://github.com/actions/runner-images/issues/1336

Probably not too much of an issue in normal operation, but it in situations like we have now where the CI queue resembles Clapham Junction, that task appears to be a bottleneck.

tobil4sk:
Could this be because it's emulating x86_64 on arm64 (because we're using macos-latest)? We could check if it's faster if we go back to macos-13 which should still be x86_64 by default.

Aidan63:
I just had a look at the results from the commit from before I moved it over to macos-latest. On macos-13 the haxe test suite takes around 15mins to run.

https://github.com/HaxeFoundation/hxcpp/actions/runs/9553228673/job/26331610744

The slow part is compiling all the C++, is the clang executable going through the emulation layer as well when it gets spawned from the hxcpp build tool (assuming clang is a native arm in the first place)?

tobil4sk:
Setup haxe installs x86_64 versions of haxe and neko, so neko is running in the x86_64 environment and I guess that means any of the processes it calls will also be x86_64.

We could fix this by installing arm64 versions of haxe and neko from homebrew, or by adding arm64 support for setup-haxe but the haxelib binary is not yet built properly for arm64 (see https://github.com/HaxeFoundation/haxe/pull/11663). Or we could revert back to macos-13 for now.

Aidan63:
Might be best to switch back to macos-13 for now, and once all the links in the chain have arm versions try again and see what the performance is like.

tobil4sk:
The nightly builds are now fully universal (https://github.com/HaxeFoundation/haxe/pull/11663 was merged)! So hopefully that might give us a speed up.

Aidan63:
Quickly bodge setup-haxe to use the new neko release candidates and building the haxe test suite has now gone done to 6mins, which is a nice speed up. I'm reasonably confident this is all using native arm versions as the cffi test is failing with an architecture mismatch error.

https://github.com/Aidan63/hxcpp/actions/runs/9799095232

Full run results for reference.

[...]

PR #1123, opened 2024-06-27
CI - Updated version handling

Aidan63:
This has all been far more CI stuff than I initially bargained for, but here's one more before I lose the will to CI or forget it all.

This is mostly a cleanup of the auto incrementing version stuff to avoid using the ci run counter and my workaround I added the other day. The logic is as follows.

- The last tag is fetched and compared against the version number in haxelib.json
- If the major and minor versions of the two match then the patch number of the tag is incremented and used as the next tag / version number
- If the haxelib.json major / minor combination is newer than the last tag then the haxelib version is used with the patch number set to zero.
- Any other version comparison results are treated as failures.

This means we still have the incrementing patch number for releases and by updating the version in haxelib.json we can reset the patch counter for new major / minor releases. E.g. changing the haxelib.json version to 5.0.x will result in the next release being 5.0.0 instead of 4.3.whatever.

I also updated some of the CI actions used since several of them were depricated and going to be disabled in November. This is a bit hard to confirm that it works, but by modifying the yml before opening this merge I was able to create releases in my fork.


Aidan63:
Seeing as one of the mac haxe test suite actions got stuck I've just downgraded it to use macos-13 (x86) to get it to run again and get a speed up. Once all parts of the haxe puzzle have a native arm version I'll revisit to see if its still really slow.

https://github.com/HaxeFoundation/hxcpp/issues/1121

Simn:
This PR now has a conflict.

Aidan63:
@Simn updated to master

PR #1125, opened 2024-06-29
[ssl] Separate mbedtls c files from SSL.cpp

tobil4sk:
#1111 fixed an error with old android ndks which used old versions of gcc that didn't set c99 by default. However, it turns out this breaks new ndks, because they complain about setting `-std=c99` when compiling a c++ file (SSL.cpp):
```
Error: error: invalid argument '-std=c99' not allowed with 'C++'
```

We can fix the issue by separating the c files from the c++ files so that SSL.cpp is compiled without this flag.


tobil4sk:
@Simn Please could this be merged? This is starting to show up as a warning with gcc now as well:

```
warning: command-line option ‘-std=c99’ is valid for C/ObjC but not for C++
```

PR #1129, opened 2024-07-11
[ssl] fix link order

Apprentice-Alchemist:


tobil4sk:
The linking order issue happens when using the compiler cache, and was introduced by #1125.

PR #1139, opened 2024-08-02
Faster CFFI Calls

NeeEoo:
Its not needed to load the function every time, so its better to make it only load once

### DIFF
--- a/include/hx/CFFILoader.h
+++ b/include/hx/CFFILoader.h
@@ -289,11 +289,13 @@ extern "C" ret name def_args;
    extern FUNC_##name name; \
    ret IMPL_##name def_args \
    { \
-      name = (FUNC_##name)LoadFunc(#name); \
-      if (!name) \
-      { \
-         fprintf(stderr,"Could not find function:" #name " \n"); \
-         exit(1); \
+      if(!name) { \
+         name = (FUNC_##name)LoadFunc(#name); \
+         if (!name) \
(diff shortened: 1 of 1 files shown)


SomeGuyWhoLovesCoding:
I'm definitely adding this to my modified hxcpp

hughsando:
I think the function should only get loaded once.
The idea is that "name" (the function you are calling) starts as function pointer, pointing to the "IMPL_" function which is actually a loader function.  So name should never theoretically start as the null pointer (barring static order initialization issues).  The loader function overwrites the "name" function pointer to point the real version, so subsequent calls will go straight to the real function and do not even need to try loading again.  I print statement should verify if this is working correctly.

NeeEoo:
> I think the function should only get loaded once. The idea is that "name" (the function you are calling) starts as function pointer, pointing to the "IMPL_" function which is actually a loader function. So name should never theoretically start as the null pointer (barring static order initialization issues). The loader function overwrites the "name" function pointer to point the real version, so subsequent calls will go straight to the real function and do not even need to try loading again. I print statement should verify if this is working correctly.

Yea that would likely work! I'll work on implementing it soon

NeeEoo:
Closing this, since i noticed it broke compiling with lime somehow. If someone figures out a fix please tell me (val_string call in Font.fromFile caused issues, probs calling from a stale reference)

PR #1140, opened 2024-08-02
Compile Progress

NeeEoo:
This changes the compile header to be

```
Compiling group: haxe (2997 files)
```

and adds a small tag saying the progress of the compilation

Skärmavbild 2024-08-02 kl  12 57 15

### DIFF
--- a/tools/hxcpp/BuildTool.hx
+++ b/tools/hxcpp/BuildTool.hx
@@ -576,7 +576,7 @@ class BuildTool
                   first = false;
                   Log.lock();
                   Log.println("");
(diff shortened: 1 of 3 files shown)


SorbetLover:
Bro is cooking

MAJigsaw77:
I think it should look better if you don't add that space when is for example [ 0.8%].

Simn:
The space is important for alignment:

```
[9.8%]
[9.9%]
[10.0%] <-- asdf
```

mcagabe19:
and it'd better if it print `1 file(s)` instead of `1 files`

PR #1143, opened 2024-08-18
Fix Callstack Retrieval in Critical Error Handler.

MAJigsaw77:
Before this fix, the critical error handler retrieved the callstack from a previous error. Now, it correctly captures the callstack from the error that triggered the crash handler.


Simn:
I'm not familiar with this code but just from the sequence of statements this makes sense to me, so I'll merge it!

PR #1144, opened 2024-08-27
Fix deprecated non-prototype warnings in Zlib for non-MSVC compilers.

MAJigsaw77:
This PR adds the `-Wno-deprecated-non-prototype` compiler flag to the Zlib build process for non-MSVC compilers. The purpose of this change is to remove warnings related to deprecated non-prototype function definitions.

### DIFF
--- a/src/hx/libs/zlib/Build.xml
+++ b/src/hx/libs/zlib/Build.xml
@@ -13,6 +13,7 @@
   <compilerflag value="-DSTDC" unless="windows" />
   <compilerflag value="-DHAVE_UNISTD_H" unless="windows" />
   <compilerflag value="-Wno-deprecated-non-prototype" unless="MSVC_VER" />
+  <compilerflag value="-Wno-unknown-warning" unless="MSVC_VER" />
 
   <file name="ZLib.cpp"/>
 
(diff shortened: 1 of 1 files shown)


MAJigsaw77:
Once `hxcpp` starts using `zlib` 1.3 this change can be removed as mentioned [here](https://github.com/madler/zlib/issues/842#issue-1853663795)

hughsando:
This flag itself causes warnings when clang is older (<15 ?).  This could be countered with `<compilerflag value="-Wno-unknown-warning-option" unless="MSVC_VER" />`   It may also need `<compilerflag value="-Wno-unknown-warning" unless="MSVC_VER" />` for gcc.  Not a bad solution, but upgrading zlib may be better.

MAJigsaw77:
> This flag itself causes warnings when clang is older (<15 ?). This could be countered with `<compilerflag value="-Wno-unknown-warning-option" unless="MSVC_VER" />` It may also need `<compilerflag value="-Wno-unknown-warning" unless="MSVC_VER" />` for gcc. Not a bad solution, but upgrading zlib may be better.

I'll add `<compilerflag value="-Wno-unknown-warning" unless="MSVC_VER" />` in any case.

hughsando:
We will see how this goes for now, and think about zlib updates.  Thanks!

PR #1145, opened 2024-08-29
Resolved multiple warnings on `Apple` targets.

MAJigsaw77:
* Replaced `sprintf` with `snprintf` because `sprintf` is deprecated.
* Added `-Wno-unused-command-line-argument` to MacOS toolchain to remove warnings related to unused commands, for example `clang: warning: argument unused during compilation: '-stdlib=libc++' [-Wunused-command-line-argument]`.

### DIFF
--- a/src/Dynamic.cpp
+++ b/src/Dynamic.cpp
@@ -217,7 +217,7 @@ public:
    String toString()
    {
       char buf[100];
-      sprintf(buf,"Pointer(%p)", mValue);
+      snprintf(buf, sizeof(buf), "Pointer(%p)", mValue);
       return String(buf);
    }
    String __ToString() const { return String(mValue); }
--- a/src/Dynamic.cpp
+++ b/src/Dynamic.cpp
@@ -217,7 +217,7 @@ public:
    String toString()
    {
       char buf[100];
-      snprintf(buf, sizeof(buf), "Pointer(%p)", mValue);
+      snprintf(buf,sizeof(buf),"Pointer(%p)", mValue);
       return String(buf);
    }
    String __ToString() const { return String(mValue); }
--- a/src/hx/CFFI.cpp
+++ b/src/hx/CFFI.cpp
@@ -106,7 +106,7 @@ public:
          return _hxcpp_toString( Dynamic(this) );
 
       char buffer[40];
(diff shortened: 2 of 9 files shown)


tobil4sk:
> Added -Wno-unused-command-line-argument to MacOS toolchain to remove warnings related to unused commands, for example clang: warning: argument unused during compilation: '-stdlib=libc++' [-Wunused-command-line-argument].

It would be better to solve the root cause of this warning, which is that `-stdlib=libstdc++` should not be passed in when compiling regular C files. It is possible to use cppflag and mmflag tags, e.g.:

https://github.com/HaxeFoundation/hxcpp/blob/17ac6e8a023a8c7a6af0dd8d41f86879e550b5b1/toolchain/iphoneos-toolchain.xml#L35

MAJigsaw77:
@tobil4sk is it fine now?

MAJigsaw77:
@tobil4sk Are the changes from `iOS` and `tvOS` toolchains fine as well?

mcagabe19 (approved):

MAJigsaw77:
> @tobil4sk Are the changes from `iOS` and `tvOS` toolchains fine as well?

Tested them, they work flawlessly

tobil4sk:
> @tobil4sk Are the changes from iOS and tvOS toolchains fine as well?

Thanks, they look good to me too

src/Dynamic.cpp
```diff
@@ -217,7 +217,7 @@ class PointerData : public hx::Object
    String toString()
    {
       char buf[100];
-      sprintf(buf,"Pointer(%p)", mValue);
+      snprintf(buf,sizeof(buf),"Pointer(%p)", mValue);
```
tobil4sk, line 220: Was this correct? `sizeof()` returns the size of a single `char`, it doesn't return `100`.

tobil4sk: Actually, it looks like in this case it does work properly since `char buf[]` is stack allocated. Is it also working correctly in `mysql/my_api.cpp`?

tobil4sk: Would seem like this is also the case there: https://github.com/HaxeFoundation/hxcpp/blob/08f88ff314eeaa165dd76272f92c8030cc879ec2/src/hx/libs/mysql/my_proto.h#L114

Should all be good then.

PR #1146, opened 2024-08-30
Better ANSI Color Support Detection

MAJigsaw77:
This PR improves how we check for ANSI color support by handling more terminals and CI environments and simplifying the detection logic in `stripColor`.

### DIFF
--- a/tools/hxcpp/BuildTool.hx
+++ b/tools/hxcpp/BuildTool.hx
@@ -1603,8 +1603,11 @@ class BuildTool
       }
 
 
-      if (defines.exists("HXCPP_NO_COLOUR") || defines.exists("HXCPP_NO_COLOR"))
+      if (Sys.getEnv("HXCPP_COLOUR") != null || Sys.getEnv("HXCPP_COLOR") != null)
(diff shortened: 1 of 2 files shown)


Sword352:
image
Can confirm it works as excepted

MAJigsaw77:
@tobil4sk May i get a review on this pr?

tobil4sk:
Is it not possible to perform a more generic check for ansi color support? It's possible that these CI environments will change, or that not all environments are covered.

MAJigsaw77:
> Is it not possible to perform a more generic check for ansi color support? It's possible that these CI environments will change, or that not all environments are covered.

This is best i could do, for CI they don't define `TERM` or anything else.

MAJigsaw77:
The best solution is to use these as of now.

tobil4sk:
What about adding a flag like `HXCPP_COLOR`, and then the scripts that install haxe in these ci environments can just make sure that is set?

MRK2009 (approved):
Looks good to me, since you added the `HXCPP_COLOUR` flag.

Issue #569, opened 2017-02-23
Socket.cpp:361:56: error: 'inet_pton' was not declared in this scope


sebthom:
When compiling the [haxe-doctest](https://github.com/vegardit/haxe-doctest/) project to C++ using Haxe 3.4 and hxcpp 3,4,49 and GCC 5.1 or 6.1 I am getting the following exception:

```
 - Socket.cpp
Error: C:/apps/dev/haxe/haxe-3.4.0/lib/hxcpp/3,4,49/src/hx/libs/std/Socket.cpp: In function 'Array<unsigned char> _hx_std_host_resolve_ipv6(String, bool)':
C:/apps/dev/haxe/haxe-3.4.0/lib/hxcpp/3,4,49/src/hx/libs/std/Socket.cpp:361:56: error: 'inet_pton' was not declared in this scope
    int ok = inet_pton(AF_INET6, host.__s, (void *)&ipv6);
                                                        ^
C:/apps/dev/haxe/haxe-3.4.0/lib/hxcpp/3,4,49/src/hx/libs/std/Socket.cpp: In function 'String _hx_std_host_to_string_ipv6(Array<unsigned char>)':
C:/apps/dev/haxe/haxe-3.4.0/lib/hxcpp/3,4,49/src/hx/libs/std/Socket.cpp:427:55: error: 'inet_ntop' was not declared in this scope
    return String( inet_ntop(AF_INET6, &ip[0], buf, 100) );
                                                       ^
Error: Build failed
```

any ideas?

sebthom:
Same with GCC6.3

hughsando:
That's pretty odd.  Its is supposed to be in <arpa/inet.h> or  <netdb.h>
I wonder if there is some "enable ipv6" flag that is missing somewhere, or if one of the other net files is using a pre-processor to prevent the header being included.
I assume this is for linux - anything odd about the compiler setup?
I've seen gcc4 (android) and clang more so that gcc5/6 I guess.

sebthom:
I am using Windows 10 and tried these GCC distributions:
GCC 5.1: https://sourceforge.net/projects/tdm-gcc/files/TDM-GCC%20Installer/
GCC 6.1: https://sourceforge.net/projects/mingw-w64/files/Multilib%20Toolchains%28Targetting%20Win32%20and%20Win64%29/ray_linn/gcc-6.X-with-ada/
[...]

Issue #574, opened 2017-03-03
Windows XP compatability issues.


Yanrishatum:
Currently HXCPP application crashes on Windows XP due to usage of functions that are not present in Window XP. Both app and libraries (std/zlib/etc) are built with `-DHXCPP_WINXP_COMPAT` flag.  
Identified at least two of them: `inet_pton` and `iten_ntop` in `winsocks2.h` / `WS2_32.dll`.
Additionally dll files that are bundled in Haxelib release is clearly broken for WinXP (`false` for `FileSystem.exists(Sys.getCwd())` is clearly something out of nightmares.)

hughsando:
I have loaded these functions dynamically - hopefully should make this link for you.
Not sure about the other issues.  Hxcpp should no longer be using dlls - the Sys.getCwd should be coming from src/hx/libs/std/Sys.cpp now.  So that might be a place to start.

Yanrishatum:
It seems that problem with MSVC runtime package. `stat` functions returns -1, which is apparently know issue. So far problem not with HXCPP directly on that problem.

Yanrishatum:
Closing issue, as `inet_*` was fixed, and other issues were not directly HXCPP-related, and required updating Windows Kit.  Thanks.

PR #1152, opened 2024-09-06
Fixed some warnings on `MinGW` target when compiling `Socket.cpp`.

MAJigsaw77:
This PR fixes the following build warnings:

- **Redefinition of `_WIN32_WINNT`**: The macro `_WIN32_WINNT` was being redefined in `Socket.cpp`. A conditional check has been added to prevent this.
  
- **Ignored `dllimport` attribute**: Warnings related to the `dllimport` attribute for `inet_pton_func` and `inet_ntop_func` were addressed by removing the unnecessary `WINSOCK_API_LINKAGE`.

### DIFF
--- a/src/hx/libs/std/Socket.cpp
+++ b/src/hx/libs/std/Socket.cpp
@@ -9,7 +9,9 @@
 
 #ifdef __GNUC__
    // Mingw / gcc on windows
+   #ifndef _WIN32_WINNT
    #define _WIN32_WINNT 0x0501
+   #endif
    #include <winsock2.h>
    #include <ws2tcpip.h>
 #else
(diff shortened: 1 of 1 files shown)


MAJigsaw77:
By looking at this issue https://github.com/HaxeFoundation/hxcpp/issues/574 maybe something might break without `WINSOCK_API_LINKAGE`.

MAJigsaw77:
@hughsando Can i get a review on this pr?

tobil4sk:
Also should be investigated whether it is still necessary to set `_WIN32_WINNT`

tobil4sk:
The unnecessary `WINSOCK_API_LINKAGE` seems to have been copied and pasted from the previous code, where the loading relied on `WINSOCK_API_LINKAGE` rather than `GetProcAddress` as it does now: https://github.com/HaxeFoundation/hxcpp/commit/b08a79d119d8646c10f713010a8b2d82f2a31690

Here is a similar example from microsoft documentation of using `GetProcAddress` (without `WINSOCK_API_LINKAGE`): https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getprocaddress#remarks

`_WIN32_WINNT` was added in [a commit](https://github.com/HaxeFoundation/hxcpp/commit/c95cbafe6ceb7bb7642dee8efcdb16fcf4d3907d) intending to fix #569. The issue links to this [stackoverflow post](https://stackoverflow.com/questions/15370033/how-to-use-inet-pton-with-the-mingw-compiler), which says:

> You may need to set _WIN32_WINNT to a code representing the minimum supported Windows version. That enables the APIs that are only supported in the more recent versions of Windows. MSVC does this for you, MinGW does not, but it is easy to fix.

Mingw-w64 now defines `_WIN32_WINNT` by default, so maybe this is no longer necessary. It is safe to keep if it is wrapped in an `#ifndef` guard as in this patch.

So these changes seem to be correct.

tobil4sk (approved):
As explained a both, these solutions to these warnings make sense and would be useful for cleaning up the compilation output with mingw.

Issue #1150, opened 2024-09-04
[cppia] Cant seem to get host to call extern super functions


ianharrigan:
So Ive reduced my problem scope to the following:

In the cppia host:
```haxe
class FakeButton extends FakeInteractiveComponent {
    public override function someFunction() {
        super.someFunction();
        trace(">>>>>>>>>>>>>>>>>>>>>>>>> SOME FUNCTION: FakeButton");
    }
}

class FakeInteractiveComponent extends FakeComponent {
    public override function someFunction() {
        super.someFunction();
        trace(">>>>>>>>>>>>>>>>>>>>>>>>> SOME FUNCTION: FakeInteractiveComponent");
    }
}

class FakeComponent extends FakeComponentBase {
    public override function someFunction() {
        super.someFunction();
        trace(">>>>>>>>>>>>>>>>>>>>>>>>> SOME FUNCTION: FakeComponent");
    }
}

class FakeComponentBase {
    public function new() {
    }

    public function someFunction() {
        trace(">>>>>>>>>>>>>>>>>>>>>>>>> SOME FUNCTION: FakeComponentBase");
    }
}
```
As expected if you create an instance of this in the host (`new FakeButton()`) and call the `someFunction()` then things trace as you would imagine:
```
src/fake/FakeComponentBase.hx:8: >>>>>>>>>>>>>>>>>>>>>>>>> SOME FUNCTION: FakeComponentBase
src/fake/FakeComponent.hx:6: >>>>>>>>>>>>>>>>>>>>>>>>> SOME FUNCTION: FakeComponent
src/fake/FakeInteractiveComponent.hx:6: >>>>>>>>>>>>>>>>>>>>>>>>> SOME FUNCTION: FakeInteractiveComponent
src/fake/FakeButton.hx:6: >>>>>>>>>>>>>>>>>>>>>>>>> SOME FUNCTION: FakeButton
```

All fine, however, in the script file, i have a set of externs for these classes, as well as a class that extends one of these externs:
```haxe
@:expose("MyFakeButton")
class MyFakeButton extends FakeButton {
    public override function someFunction() {
        super.someFunction();
        trace(">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> MyFakeButton.someFunction");
    }
}

extern class FakeButton extends FakeInteractiveComponent {
    public function someFunction():Void;
}

[...]

PR #1154, opened 2024-09-15
add cppia test for multi level inheritance in host

Aidan63:
Adds tests for https://github.com/HaxeFoundation/hxcpp/issues/1150

haxe side here https://github.com/HaxeFoundation/haxe/pull/11773


Aidan63:
I can't remember, but going by the name I think it was supposed to be in the host.

PR #1157, opened 2024-10-14
Fix sys stat size integer overflow

barisyild:
This pull request fixes the file size integer overflow issue for sys_stat.

### DIFF
--- a/src/hx/libs/std/Sys.cpp
+++ b/src/hx/libs/std/Sys.cpp
@@ -415,7 +415,8 @@ void  _hx_std_sys_rename( String path, String newname )
       hx::Throw(HX_CSTRING("Could not rename"));
 }
 
-#define STATF(f) o->Add(HX_CSTRING(#f),(int)(s.st_##f))
+#define STATF32(f) o->Add(HX_CSTRING(#f),(int)(s.st_##f))
+#define STATF64(f) o->Add(HX_CSTRING(#f),(long)(s.st_##f))
 
 /**
    sys_stat : string -> {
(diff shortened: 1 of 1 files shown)


Simn:
CI failure is related:

```
Error: /Users/runner/work/hxcpp/hxcpp/src/hx/libs/std/Sys.cpp:479:4: error: conversion from 'long' to 'const Dynamic' is ambiguous
   STATF64(size);
```

barisyild:
> CI failure is related:
> 
> ```
> Error: /Users/runner/work/hxcpp/hxcpp/src/hx/libs/std/Sys.cpp:479:4: error: conversion from 'long' to 'const Dynamic' is ambiguous
>    STATF64(size);
> ```

I think Apple is a bit more strict with the type casting, the compile problem was solved.

https://github.com/barisyild/hxcpp/actions/runs/11325731311

Apprentice-Alchemist:
Note that `long` is 32-bit on certain platforms, it would be better to use `int64_t`.

Simn:
Can't we just always use the `__APPLE__` version?

Aidan63:
Doesn't this require a change on the haxe side to work properly? Since the `FileStat` typedef only has size as an Int.

Changes like this should also probably be guarded with a `HXCPP_API_LEVEL` check so new hxcpp versions can be used with older haxe without anything changing.

barisyild:
> Doesn't this require a change on the haxe side to work properly? Since the `FileStat` typedef only has size as an Int.
> 
> Changes like this should also probably be guarded with a `HXCPP_API_LEVEL` check so new hxcpp versions can be used with older haxe without anything changing.

I tried and you are right, I think it needs a haxe api change.

barisyild:
https://github.com/HaxeFoundation/haxe/pull/11791

SomeGuyWhoLovesCoding:
I just tried it and it doesn't work. FileStat.size just returns 0.

PR #1165, opened 2024-12-07
Don't include xlocale.h on Linux without glibc

hannesbraun:
For glibc based systems, `xlocale.h` is already excluded. On Linux distributions with musl libc (for example Alpine Linux), `xlocale.h` is unavailable as well. As there is no macro to detect musl, my suggested fix would be to rely on `__linux__` and assume that `xlocale.h` is not required on Linux systems in general.

### DIFF
--- a/src/hx/libs/std/Sys.cpp
+++ b/src/hx/libs/std/Sys.cpp
@@ -30,7 +30,7 @@
    #include <limits.h>
    #ifndef ANDROID
       #include <locale.h>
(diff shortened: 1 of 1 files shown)


tobil4sk:
These were the patch notes when xlocale.h was removed: https://sourceware.org/glibc/wiki/Release/2.26#Removal_of_.27xlocale.h.27

> We know that a number of programs do include it, but because it has never been part of any other C library, programs that use it are probably testing for its existence with autoconf or a similar tool, and should not fail to compile.

PR #1166, opened 2024-12-10
Fixed typos

NeeEoo:
Also added -nocolour due to HXCPP_NO_COLOUR

If there are any typo's you would want me to revert, please do mention below or in a code review, or push a commit to revert that typo

Edit:
i noticed that HXCPP_NO_WASM_EXEPTIONS has a typo
and \_hx\_**reslove**\_virtual_array also

### DIFF
--- a/include/Array.h
+++ b/include/Array.h
@@ -387,7 +387,7 @@ namespace cpp
 
 // --- Array_obj ------------------------------------------------------------------
 //
-// The Array_obj specializes the ArrayBase, adding typing where required
+// The Array_obj specialises the ArrayBase, adding typing where required
 
 
 namespace hx
--- a/include/Array.h
+++ b/include/Array.h
@@ -387,7 +387,7 @@ namespace cpp
 
 // --- Array_obj ------------------------------------------------------------------
 //
-// The Array_obj specialises the ArrayBase, adding typing where required
+// The Array_obj specializes the ArrayBase, adding typing where required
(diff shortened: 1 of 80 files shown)


include/Array.h
```diff
@@ -387,7 +387,7 @@ namespace cpp
 
 // --- Array_obj ------------------------------------------------------------------
 //
-// The Array_obj specialises the ArrayBase, adding typing where required
+// The Array_obj specializes the ArrayBase, adding typing where required
```
Simn, line 390: This one was fine, we can BE instead of AE.

include/hx/StdString.h
```diff
@@ -1,5 +1,5 @@
-#ifndef HX_STD_STRING_INCLUDEDED
```
Simn, line 1: lol...

Simn:
Wow... It's scary how many typos there are. Thanks a lot for looking into this!

include/Array.h
NeeEoo: Alright ill revert it while i fix more typos

NeeEoo:
> Do we want to fix typos in the sljit_src folder? That's just vendored sljit source like the other libs in the project/thirdparty folder.

i mean its not included in the .gitattributes as a linguist-vendored so i didnt think of that

MRK2009:
This is stale due to inactivity.

PR #1179, opened 2025-01-23
Add warning if destination flag is ignored

tobil4sk:
If the current target does not produce a file, the -Ddestination flag is currently ignored without warning.

This situation may occur commonly when 'default' points to another target, but doesn't produce a file itself. A warning will now be displayed:

```
Warning: Target 'default' does not output a file, so 'destination' has been ignored
```

### DIFF
--- a/tools/hxcpp/BuildTool.hx
+++ b/tools/hxcpp/BuildTool.hx
@@ -771,6 +771,8 @@ class BuildTool
                }
             }
             Profile.pop();
+         case _ if (inDestination != null):
+            Log.warn('Target \'${inTarget}\' does not output a file, so \'destination\' has been ignored');
       }
 
       if (mCopyFiles.length>0)


Simn:
I'm having a hard time making sense of that sentence, but I suppose that's because I lack context.

tobil4sk:
Suppose you have a Build.xml like this:

```xml
<xml>

	<files id="main">
		<file name="main.c" />
	</files>

	<target id="main" output="main${EXE}" tool="linker" toolid="exe">
		<files id="main" />
	</target>

	<target id="default">
		<target id="main" />
	</target>

</xml>
```

The file has two targets, `main` which compiles a file, and `default` which just runs the `main` target  (Haxe cpp gen produces a `default` and `haxe` target like this). `-Ddestination` is used to copy the output of a single target to a given file path.

Right now, if you use the `default` target (which hxcpp uses by default), the `-Ddestination` flag does not do anything since `default` does not produce an output file. The warning is meant to notify the user of this.

PR #1188, opened 2025-02-02
[cppia] Respect jit exceptions in interp mode

### LINKED ISSUE #876: Cppia - throw inside while
https://github.com/HaxeFoundation/haxe/issues/8915

tobil4sk:
When running in interp mode, native c++ exception handling is used, but when running jit mode, jumps are used along with ctx->exception.

This means that an exception thrown from jit mode is not respected in interp mode. This can cause problems with local functions (which are interpreted even in jit mode). When a local function calls a compiled function that throws an exception, the exception is ignored until we return back into a compiled function.

This patch fixes the problem by checking ctx->exception when running in interp mode, to make sure that no exception has been thrown from a jit compiled function.

This fixes #876.

### DIFF
--- a/src/hx/cppia/Cppia.cpp
+++ b/src/hx/cppia/Cppia.cpp
@@ -6346,30 +6346,38 @@ struct TryExpr : public CppiaVoidExpr
       }
       return this;
    }
+
+   void handleException(CppiaCtx* ctx, Dynamic caught) {
+      //Class cls = caught.mPtr ? caught->__GetClass() : 0;
+      for(int i=0;i<catchCount;i++)
+      {
+         Catch &c = catches[i];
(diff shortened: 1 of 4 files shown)


Simn:
@hughsando Could you check if this makes sense?

hughsando:
Yes, I think it makes sense.
I wonder if adding ctx->exception to the ctx->breakContReturn flags might make it easier to check consistently, ie ctx->breakContinueReturnException .  It may also complicate things too much - it's been a while since I've looking into this.

Simn:
Thanks for checking! I'll merge this as-is in order to allow https://github.com/HaxeFoundation/haxe/pull/11922. We can perhaps improve on it later.

PR #1189, opened 2025-02-06
Marshalling Types

Aidan63:
See for more details


kevinresol:
Apparently the array access of StdVector extern does not work:

```
error: type '::cpp::marshal::ValueType< ::std::vector<const ::godot::Variant *>>' does not provide a subscript operator
 1024 | HXLINE( 141)            ptrs[0] = ...;
      |                         ~~~~^~
```

Aidan63:
Hmm, I'm sure it worked at some point. I'll take a look.

PR #1197, opened 2025-03-05
Test for nightly cppia interface issue

Aidan63:
https://github.com/HaxeFoundation/haxe/pull/12042

Cppia test suite also now use utest.


Aidan63:
@Simn can you re-run the CI? Failing nightly tests should pass now that the haxe side has been merged in.

PR #1199, opened 2025-03-12
line buffered cpp stdout

cyn0x8:
attempts to solve (at least the cpp part of) https://github.com/HaxeFoundation/haxe/issues/8830

- changes unbuffered stdout to be line-buffered
- flushes after each `trace`/`println`

<details><summary>examples</summary><br>

its noticeably faster on my end!

https://github.com/user-attachments/assets/bcb770ca-1ee4-4e59-9e77-1e0847def3bd

---

and the test case from https://github.com/HaxeFoundation/haxe/issues/8830

```haxe
class TraceTest {

    public static function main() {
        var start = haxe.Timer.stamp();
        for (i in 1...2000)
            trace(i);
        var duration =  Math.round((haxe.Timer.stamp() - start)*100) / 100;
        trace('Duration: $duration seconds.');
    }
}
```

image

</details>

---

im new to contributing to this repo and kinda new to c++ in general so criticism or suggestions are very much appreciated

### DIFF
--- a/src/hx/StdLibs.cpp
+++ b/src/hx/StdLibs.cpp
@@ -282,7 +282,7 @@ void __hxcpp_stdlibs_boot()
    //  It does not cause fread to return immediately - as perhaps desired.
    //  But it does cause some new-line characters to be lost.
    //setbuf(stdin, 0);
-   setbuf(stdout, 0);
+   setvbuf(stdout, NULL, _IOLBF, BUFSIZ);
    setbuf(stderr, 0);
 }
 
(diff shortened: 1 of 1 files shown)


EliteMasterEric:
Fascinating! The test case running almost 10x faster than any other language is crazy. Hope this gets approved.

AbnormalPoof (approved):
Traces are WAY faster with this change now, nice work!

Simn:
We probably don't have many trace tests, but I'll assume that if it compiles it works.

@yuxiaomao I'm not familiar with the HL implementation, but perhaps something similar could be done there?

PR #992, opened 2022-05-24
Support Android NDKs higher than 21

player-03:
Version 22 changed the NDK's structure, meaning some files and folders aren't where hxcpp expects them.

---

To test this, create the following three files, and compile with `haxelib run hxcpp Build.xml -Dandroid`.

1. Hello.cpp:

   ```cpp
   #include <stdio.h>
   
   #include <atomic>
   
   int main(int argc, char **argv) {
   	printf("Hello world\n");
   	return 0;
   }
   ```

2. Build.xml:

   ```xml
   <xml>
   	<set name="HXCPP_ARM64" value="1" if="android" />
   	
   	<files id="hello">
   		<file name="Hello.cpp" />
   	</files>
   	
   	<target id="default" output="hello" tool="linker" toolid="${STD_MODULE_LINK}">
   		<outdir name="bin" />
   		<ext value="" />
   		<files id="hello" />
   	</target>
   </xml>
   ```

3. Options.txt: (empty)

---

In my tests, NDK r21e worked fine, but r23 (edit: and r24) ran into multiple errors upon reaching `#include <atomic>`. (First it couldn't find the file, then it couldn't compile it.)

After applying the changes in this pull request, r21e, r23, and r24 all work correctly. I'm not certain there are no regressions elsewhere, but I tried to minimize that risk.

Potentially useful reference: https://developer.android.com/ndk/guides/other_build_systems

---

Closes #706.

PR #1201, opened 2025-03-13
Add support for HXCPP_ANDROID_PLATFORM define

tobil4sk:
Originally, android min sdk version could be configured using `PLATFORM=android-xx`. This was used because it was a sub directory of the android ndk: 9e716f16735bd79fc39729f4334a73a393399427. Hxcpp used to extract the version from this define: https://github.com/HaxeFoundation/hxcpp/blob/75108a59f44c69348782f0be6204da512ebe1397/tools/hxcpp/Setup.hx#L547-L552

However, for modern ndks, PLATFORM is no longer used to resolve any directory and since #825 it is not parsed by hxcpp for ndk 20 or newer either.

This left people with no way to configure the min sdk version. To give a way of doing this again, `PLATFORM_NUMBER` was made configurable in #992. However, given that it does not have a very descriptive name this seems like it was an internal variable that should not have been exposed like this. It has also never been documented.

This PR allows setting the minimum sdk version using `HXCPP_ANDROID_PLATFORM`. This is a flag that was already used internally by hxcpp and has a much more descriptive name.

I have therefore also deprecated the PLATFORM_NUMBER define with a warning.

### DIFF
--- a/docs/build_xml/Defines.md
+++ b/docs/build_xml/Defines.md
@@ -67,7 +67,7 @@ Defines affecting target architecture.
 | *HXCPP_LINUX_ARM64*     | Run on a linux ARM64 device |
 | *winrt*                 | Compile for windowsRt/windows UWP |
 | *android*               | Compile for android |
-| *PLATFORM*              | Specify the android platform for NDK compilation |
+| *HXCPP_ANDROID_PLATFORM* | Specify the android platform for NDK compilation |
 | *ANDROID_NDK_ROOT*      | Specify the location of the android NDK toolchain |
 | *ANDROID_NDK_DIR*       | Specify the search location for finding the android NDK toolchain |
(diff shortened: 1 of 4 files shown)


tobil4sk:
Actually, instead hxcpp manually specifying which api version to use per ndk version, it should read the ndk's metadata to determine this, which avoids having to bump it every time and is closer to what it used to do with older ndks when it would read the `platforms/` folder contents

tobil4sk:
The default `HXCPP_ANDROID_PLATFORM` value is now set based on the ndk metadata, which means it won't be necessary in future to have PRs like: #1134 (or the platform version bump in #992).

Later it may be a good idea if the branch for ndks older than 20 would be unified with this, because right now we use the maximum version there, but the minimum version for newer ndks. However, we need to use the minimum platform version that supports the given architecture.

PR #1208, opened 2025-03-27
Fix android build with ndk 19

tobil4sk:
Like r20, r19 had the preconfigured clang binaries/scripts, however, it still required explicitly passing `-stdlib=libc++`. See: https://android.googlesource.com/platform/ndk/+/refs/tags/ndk-r19/docs/BuildSystemMaintainers.md#required-android-specific-arguments


Simn:
Need an update.

tobil4sk:
Thanks! updated

PR #1212, opened 2025-04-10
[cppia] Fix missing jit exception checks

tobil4sk:
Even with #1188, some checks for jit exceptions in interp code were still missing. This causes control flow bugs when local functions interact with other code, as they are currently not jit compiled.

In short, we have to make sure an exception hasn't been thrown after evaluating the object for a method call.


Simn:
@hughsando Please check!

tobil4sk:
> @hughsando Please check!

It would be worth double checking if the full `BCR_CHECK`s are required in all cases, perhaps in some places it's enough just to check only `ctx->exception`.

`CallFunExpr` has the full check [after getting `this`](https://github.com/HaxeFoundation/hxcpp/blob/00985d5d768c1fac8a0f8b85d2bd52d4f2dea559/src/hx/cppia/Cppia.cpp#L768), and `CallDynamicFunction` has it [after each argument](https://github.com/HaxeFoundation/hxcpp/blob/00985d5d768c1fac8a0f8b85d2bd52d4f2dea559/src/hx/cppia/Cppia.cpp#L959), so I based my changes off those two. Perhaps there was a genuine reason why `Call`, `CallHaxe` and `CallMemberVTable` might not require the BCR check everywhere.

Simn:
Hmm, [this run](https://github.com/HaxeFoundation/hxcpp/actions/runs/22340505784/job/64642538991) seems to have had a weird state:

```
Run haxe compile-cpp.hxml -D HXCPP_M64 -D no_http
Error:  src/unit/TestCoroutines.hx:152: characters 29-52

 152 |  public function iterator():haxe.coro.AsyncIterator<Int> {
     |                             ^^^^^^^^^^^^^^^^^^^^^^^
     | Type not found : haxe.coro.AsyncIterator
```

This must mean that tests and std are not in-sync. Unrelated to this PR but something to look into because that can lead to all sorts of strange errors.

tobil4sk:
Yeah, it looks like it just clones the latest haxe repository even if the haxe from build.haxe.org is slightly older. I guess we should read the commit hash from the version and use that, which is what the azure script used to do:
https://github.com/HaxeFoundation/hxcpp/blob/8e3bc36641b3c28610ecc621204d8bf64fc25ed6/tools/azure-pipelines/build.yml#L145

PR #1214, opened 2025-04-14
Support unsigned long values for Dynamic

singpolyma:
Fixes ambiguous calls when compiling with Xcode


barisyild:
> why would it break compatibility? anything using an unsigned long (or equivalent) before would fail to compile

Isn't this change likely to affect the behavior of existing projects?

singpolyma:
It shouldn't affect the behaviour of anything existing, except that some that would not compile before will compile now.

PR #1220, opened 2025-04-26
Custom stack traces

Aidan63:
See https://github.com/HaxeFoundation/haxe/pull/12213

This updates hxcpp unhandled exceptions to only call `__hx_dump_stack` if the caught object does not have a `__customStack` field, in which case that is printed out instead.

### DIFF
--- a/include/hx/HxcppMain.h
+++ b/include/hx/HxcppMain.h
@@ -97,7 +97,7 @@
           auto customStack = e->__Field(HX_CSTRING("__customStack"), HX_PROP_DYNAMIC).asString();
           if (::hx::IsNotNull(customStack))
           {
-              printf("%s\n", customStack.c_str());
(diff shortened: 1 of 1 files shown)


Simn:
Hmm... the HL implementation is stricter in the sense that it actually checks if we're looking at an instance of `haxe.Exception`. Though now I wonder if that's necessary. I'm generally wary of magic field names like this but *if* we add one I really think it should be something leading with `_hx_`. We don't really reserve anything like `__customStack` in general, so if someone had such a field with a different type we run into trouble here.

Aidan63:
Good point, I've changed it to `_hx_customStack` the field on the cpp haxe exception class will need updating. I'm going to bump up looking at revamping hxcpps exception stuff on my internal todo list, so hopefully it won't be around for too long.

PR #1232, opened 2025-06-01
add support for cpp20

dazKind:


MRK2009:
conflicted

Aidan63:
I'm happy with the cpp changes, but I don't think we want to add any more cpp version defines. We've since added a `HXCPP_CXX_STANDARD` define (e.g. `-D HXCPP_CXX_STANDARD=20`) to specify the version and the ability to specify the  version on individual `files` elements inside build xml.

MRK2009:
> I'm happy with the cpp changes, but I don't think we want to add any more cpp version defines. We've since added a `HXCPP_CXX_STANDARD` define (e.g. `-D HXCPP_CXX_STANDARD=20`) to specify the version and the ability to specify the version on individual `files` elements inside build xml.

I mean they could've just done it in a alternative way by adding it in the `BuildTool` class instead of this.

Issue #570, opened 2017-02-23
[ios] SSL - No CA Chain is set, but required to operate


kevinresol:
I think this is because _hx_ssl_cert_load_defaults isn't implemented on ios.

hughsando:
I'm not sure where to get the root certificate from on ios.
One option might be to disable the check with Socket.DEFAULT_VERIFY_CERT = false, but this might introduce security issues.

kevinresol:
I think there is no way to read root certs in iOS due to its sandboxed nature.

I think hxcpp can use CFNetwork (available as c++) for SSL connections

hughsando:
Yes, a native solution might be the best way.  hxcpp can inter-op with objc reasonably well now, so a haxe-only solution (no external source code) might be possible.  Pull-requests welcome.

You could also ship your own root certificates, and read them from the assets area.

kevinresol:
> Pull-requests welcome.

I really wished to. But I have totally no idea about c/c++.
Maybe I really have to learn that at some point.

alanlanglois:
Can it be related?
https://github.com/yupswing/akifox-asynchttp/issues/43

PR #1239, opened 2025-07-08
Stub in new SSL implemenation for Apple OS

singpolyma:
Use native SSL on Apple OS instead of mbedtls. Especially needed on iOS where the root CA store cannot be read.

Fixes https://github.com/HaxeFoundation/hxcpp/issues/570

### DIFF
--- a/src/hx/libs/ssl/SSL.cpp
+++ b/src/hx/libs/ssl/SSL.cpp
@@ -1,3 +1,223 @@
+#if defined(HX_MACOS) || defined(IPHONE) || defined(APPLETV)
+
+#include <Security/Security.h>
+#include <Security/SecureTransport.h>
+#include <CoreFoundation/CoreFoundation.h>
+#include <unistd.h>
+#include <sys/socket.h>
+#include <errno.h>
+
+#include <hxcpp.h>
+
+#define val_ssl(o)    ((sslctx*)o.mPtr)
+#define val_conf(o)   ((sslconf*)o.mPtr)
+#define val_socket(o) ((SocketWrapper*)o.mPtr)
+#define val_cert(o)   ((sslcert*)o.mPtr)
+#define val_pkey(o)   ((sslpkey*)o.mPtr)
+
+struct SocketWrapper : public hx::Object {
+	HX_IS_INSTANCE_OF enum { _hx_ClassId = hx::clsIdSocket };
+	int socket;
+};
+
+struct sslcert : public hx::Object {
+	HX_IS_INSTANCE_OF enum { _hx_ClassId = hx::clsIdSslCert };
+	void create() { _hx_set_finalizer(this, finalize); }
+	void destroy() {}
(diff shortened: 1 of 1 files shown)


Aidan63:
I think using OS SSL libraries where possible is a good idea (I've got a pending merge in #1135 which does it for Windows) and have looked at this mac library before, I've got a few comments though.

- Secure Transport is deprecated and it is now recommended to use the Network Framework instead, this is important as Secure Transport does not support TLS 1.3. Hxcpp doesn't at the moment either but the pending mbedtls update and the Windows SChannel merges would add TLS 1.3 support, so this would be a step backwards for mac users.
- Do Apply have cryptographic libraries which can be used to implement all the hashing and digest functions? Would be nice to use those if possible as it would allow mac builds to not need mbedtls at all.
- Could we have a flag to switch back to mbedtls for mac? In #1135 I have it set to use SChannel by default but the `HXCPP_USE_MBEDTLS` flag allows you to go back to mbedtls.

barisyild:
I think best solution is using native http api of apple os.

The sample code is a nme wrapper, so it requires a few changes to compile.
Below is an example of a working haxe bridge, @:objc is required.
```haxe
#if ios
private function iOSResponse(data:NSData, response:NSURLResponse, error:NSError)
{
	var statusCode = -1;
	if (response != null) {
		var httpResponse:NSHTTPURLResponse = untyped __cpp__("(NSHTTPURLResponse *) {0}", response);
		statusCode = httpResponse.statusCode;
	}

	var shouldRetry = (error != null) || (statusCode >= 500);

	if (shouldRetry && retryCount < maxRetries)
	{
		retryCount++;
		Sys.println('${urlRequest.url} request failed, retrying... Attempt ${retryCount} status code ${statusCode}');
		runiOS();
		return;
	}

	if (error != null || statusCode >= 400)
	{
		onError(error != null ? error.localizedDescription.toString() : '${statusCode}');
		return;
	}

	onStatus(statusCode);
	retryCount = 0;

	var bytes = data.toBytes();
	if (urlLoader.dataFormat== URLLoaderDataFormat.BINARY)
	{
		byteData = ByteArray.fromBytes(bytes);
	}
	else
	{
		stringData = bytes.getString(0, bytes.length, UTF8);
	}

	state = URLLoader.urlComplete;
}

public function runiOS() {
	var url:NSURL = NSURL.URLWithString(this.urlRequest.url);
	var request:NSMutableURLRequest = NSMutableURLRequest.requestWithURL(url);

	for(header in headers.keys())
	{
		var value:String = Std.string(headers.get(header));
		request.setValueForHTTPHeaderField(value, header);
	}

	var session:NSURLSession = NSURLSession.sharedSession();
	var task:NSURLSessionDataTask = session.dataTaskWithRequestCompletionHandler(request, untyped __cpp__('^(NSData *data, NSURLResponse *response, NSError *error) {
		dispatch_async(dispatch_get_main_queue(), ^{
			{0}(data, response, error);
		});
	}', iOSResponse));

	task.resume();
}
#end
```

PR #1240, opened 2025-07-09
fixes multithread marking

sibest:
Multithreaded marking wasn’t working because the following preprocessor block was never executed (at least when compiling with MSVC):

#if (MAX_GC_THREADS>1)
   // You can uncomment this for better call stacks if it crashes while collecting
   #define HX_MULTI_THREAD_MARKING
#endif


hughsando:
Interesting and a little concerning.  Looks like it has not been multi-thread collecting for a few years.
Your fix looks like it restores the intention of the code.

PR #1246, opened 2025-07-21
Unconditionally enable HXCPP_ALIGN_ALLOC.

Apprentice-Alchemist:
This is required to be able to safely use `std::atomic` inside hxcpp-allocated objects.


Simn:
@hughsando Could you check if this is good?

PR #1248, opened 2025-08-01
Use `16KB` max page size on Android.

MAJigsaw77:
This pr adds support for 16 KB page sizes on both Android toolchains.

https://developer.android.com/guide/practices/page-sizes

> Starting November 1st, 2025, all new apps and updates to existing apps submitted to Google Play and targeting Android 15+ devices must support 16 KB page sizes on 64-bit devices.


tobil4sk:
I think we should add this conditionally based on ndk version, since we don't have to enable it manually for r28+. We also need an additional flag for ndk 22 and below:

https://developer.android.com/guide/practices/page-sizes#compile-16-kb-alignment

Also, does it make sense to enable it for the gcc toolchain? It is only used for ndks older than r19.

PR #1252, opened 2025-08-22
[fix] Socket: Fix stale indexes and surface close/error reliably in _hx_std_socket_poll_events

dimensionscape:
This PR contains two related fixes in `_hx_std_socket_poll_events` to make
polling safer and more consistent:

1. **Clear ridx/widx on entry and early return**  
   Prevents stale values when polling with empty registries or when
   select()/poll() errors out. Previously this could result in phantom
   indexes or poll([]) returning [null].

2. **Surface close/error conditions reliably**  
   - **Windows**: now pass and handle an `exceptfds` set to detect errors
     and connection resets. On `SOCKET_ERROR`, all current read fds are
     surfaced so userland can clean up.
   - **POSIX**: include `POLLERR | POLLNVAL` alongside `POLLIN/HUP` and
     `POLLOUT/HUP` so teardown/error conditions are not missed.

Together these changes ensure that:
- Empty registries return no ready sockets.
- Closed/reset sockets always show up as "ready" for one tick so
  userland can detect EOF and deregister them.
- The behavior is consistent across Windows and POSIX backends.


dimensionscape:
This is actually a kind of important fix for me personally as my framework is broken without it. If it's helpful, I can demonstrate a sample case that it resolves. @Simn 

But, for simplicity, I think we can agree that `poll([])` should not be returning `[null]` and that is a bug.

dimensionscape:
Finally got around to adding a test for this.

PR #1267, opened 2025-11-05
Fix `ranlib` when `libdir` is set.

MAJigsaw77:


tools/hxcpp/Linker.hx
```diff
@@ -314,6 +315,7 @@ class Linker
          {
             sys.io.File.copy( mLibDir+"/"+file_name, out_name );
             FileSystem.deleteFile( mLibDir+"/"+file_name );
+            Log.info("\x1b[1mMove file: \x1b[0m" + mLibDir+"/"+file_name + " to " + out_name);
```
tobil4sk: I think it's better to print before doing the action, that way if the action fails and crashes then we can see what stage the build was at before failing. This is also what ProcessManager.runCommand does.

PR #1273, opened 2025-11-28
[cppia] Fix haxe overridden scriptable methods

tobil4sk:
This enables an alternate solution for these bugs:
https://github.com/HaxeFoundation/haxe/issues/12374
https://github.com/HaxeFoundation/haxe/issues/12376

With this PR, haxe is able to mark methods as overrides so that they don't add an additional slot to the vtable. This way, haxe can still include them in `__scriptableFunctions` so that they can be resolved correctly in super.method calls, avoiding #1150.

Alternatively, if the haxe version is old and does not mark them, then fallback logic is used to infer overrides from matching method names. This has a runtime cost but avoids incorrect behaviour for current haxe. This only really affects 4.3.7 as older versions suffered from #1150.

### DIFF
--- a/include/hx/Scriptable.h
+++ b/include/hx/Scriptable.h
@@ -20,7 +20,7 @@ struct ScriptNamedFunction : public ScriptFunction
    ScriptNamedFunction(const ScriptFunction &s) : ScriptFunction(s), name(0), isStatic(false), superExecute(0) { }
 
 #if (HXCPP_API_LEVEL >= 500)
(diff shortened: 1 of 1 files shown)


Simn:
> `__scriptableFunctions` is generated by gencpp (e.g. [here](https://github.com/HaxeFoundation/haxe/blob/095f6d3484c3f7a7a3897dd7280b29c1bcc69dbc/src/generators/cpp/gen/cppGenClassImplementation.ml#L1041-L1043)) for each class/interface

Aah, that's the piece I was missing, thanks!

PR #1276, opened 2025-12-13
int64 over size_t for views

Aidan63:
Seems like there were two problems here.

One is something only msvc flagged up, the `get` and `set` functions on the value type reference has an ambigious conversion if the function it calls accepts `int64_t`.

The other is that the type apple clang uses for size_t doesn't have a dynamic conversion, so I've just switched the last few size_t references to int64_t (I'll also update the haxe extern).

### DIFF
--- a/include/cpp/marshal/Definitions.inc
+++ b/include/cpp/marshal/Definitions.inc
@@ -129,10 +129,10 @@ namespace cpp
             bool operator!=(const ValueReference<T>& inRHS) const;
 
             template<class K>
-            K get(int index);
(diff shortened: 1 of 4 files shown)


Aidan63:
Hmm, that is odd. I just did a HXCPP_M32 build of the haxe test suite locally in WSL and that passed fine. Those failing tests don't indicate anything either, just one of many `charCodeAt` and bytes `get` tests.

32bit linux haxe tests were also passed fine with the initial merge.

PR #1277, opened 2025-12-14
Typed zip

Aidan63:
Haxe side : https://github.com/HaxeFoundation/haxe/pull/12453

There are a few changes here.

"include/hx/zip" contains the hxcpp zip public api headers, "src/hx/zip/zlib" contains the zlib implementation of this api.

The "src/hx/zip/Build.xml" file will select the zlib implementation unless "HXCPP_ZIP_XML" is defined in which case the xml in that file is used instead. This allows providing custom implementations of the zip api without needing to fork hxcpp or shadow the haxe std library types to patch the build xml meta.

I've removed the 500 api level guards from the new View type and updated the old `_hx_inflate` and `_hx_deflate` c functions to call the new zip api. This allows us to have haxe 4 compatibility without needing to keep around the old implementation.

I've changed calling `close` multiple times to just be a no-op instead of throwing, I think this makes more sense.

### DIFF
--- a/src/hx/zip/zlib/Build.xml
+++ b/src/hx/zip/zlib/Build.xml
@@ -45,8 +45,8 @@
   </files>
 
   <target id="haxe">
-    <files id="zlib_sources" unless="HXCPP_LINK_NO_ZLIB"/>
     <files id="hxcpp_zlib_impl" />
+    <files id="zlib_sources" unless="HXCPP_LINK_NO_ZLIB"/>
   </target>
 
 </xml>


Aidan63:
I'm not sure either. I branched this off the value type branch a while ago but merged master in today, maybe because that original value type branch has now been deleted?

Aidan63:
I tried a rebase on master, but at the end of it I still had 80 odd commits in the branch.

Aidan63:
That is one thing I'm not too sure about.

I quite like having it under "src/hx/zip/zlib" since it matches more closely matches the include folder structure (I was thinking about having a "/zlib" folder in the include one as well). But it is different from whats done currently.

Another thing I want to do (which doesn't apply to the zip implementation) is reduce the number of ifdef conditions. Some files (such as Process.cpp) have loads of conditionals since it's really a posix and win32 implementation crammed into one file, it makes things really hard to follow.
For these cases I was thinking of splitting it out into separate files, but I'm not yet sure on how to structure that (dedicated /win32 and /posix folders in libs/std? Process.win32.cpp and Process.posix.cpp files?).

So part of the zip folder change was me seeing how I like it as a structure (I'm still not sure).

PR #1298, opened 2026-01-24
C++11 std mutex and condition variable

Aidan63:
Replaces the old implementation which had a bunch of ifdefs depending on the platform. Now we have a single implementation using the C++ `std` types. I would have done the same for `Lock` but for some reason it took until C++20 to get a semaphore type...

I have broken the definitions out into their own headers and exposed the classes so on the haxe side we can remove the untyped code.


Simn:
Let me know when this is good to merge, I'd like to get hxcoro back to hxcpp master soon to see if we benefit from the pch change!

PR #1303, opened 2026-02-04
An attempt at making hxcpp.h parse faster

Aidan63:
Clang compiler timings do show a ~1s reduction in parse time but this doesn't seem to have had any effect on CI times so I guess it's limited else where now.

The main change is making the two templated types in FieldRef.h non-templated by using `hx::Val` instead as this header was the biggest culprit time wise.

Given that this doesn't really improve things over all I'm not sure we want these changes, I'll keep it open for a while though in case I get any other ideas.


Aidan63:
I've reverted the `Val` change since it doesn't seem to make any difference with pch, so all that's left is moving the sorter class out of the header, which is more of a code tidy up than anything so I'll still merge it.

Issue #1294, opened 2026-01-18
Unify string encoding


Aidan63:
The UTF8 and UTF16 classes in `cpp::encoding` mostly duplicates the encoding which also occurs in `String.cpp`. I would have updated string to also use those dedicated encoding classes but doing some benchmarking of converting 5 paragraphs of lorem ipsum with a few emojis in it from UTF16 to UTF8 showed it to be twice as slow.

So there probably needs to be an optimisation pass to close the distance before unifying the two.

PR #1306, opened 2026-02-09
Optimise new encoding functions

tobil4sk:
Hopefully this counteracts the performance regression seen after they started being used for haxe.io.Bytes, see: https://github.com/HaxeFoundation/hxcpp/issues/1294#issuecomment-3807346788

This is lighter than #1302 since it doesn't introduce an external library. It also addresses the problem from a different angle, by avoiding inefficiencies in the surrounding code, rather than making the encoding/decoding itself as fast as possible. Even if we decide to use simdutf I think it would be worth making some of these changes anyway, as some of them would also benefit the simdutf implementation.

These are the changes:
- Merge encodings into a single file - having the encodings in separate files made it impossible for the compiler to inline simple function calls without link time optimisation (which is off by default).
- Avoid duplicate string iteration for ascii checks - we can perform one iteration to count length and check for ascii.
- Add utf8 encode function that allocates output - current haxe has to iterate to get the utf8 length, then we also iterate to ensure the utf8 length haxe gave is correct, which is wasteful.
- Pass char32_t by value instead of by reference - not sure if this makes a big difference but it is better practice anyway.
- Estimate utf8 length instead of iterating - this removes the need for iterating to find the length. Instead, we overallocate and truncate when done.

With this I see an improvement locally in the dox benchmark compared to master (1.3% vs 5.5% time spent in Bytes.ofString).

<details>
<summary>
Expand for flamegraphs
</summary>

Download as svg and use search bar to search for "Bytes_obj::ofString".

Unpatched:

unpatched
Patched:
patched

</details>


Aidan63:
Just did a quick bit of benchmarking with a "lorem ipsum" like file full of chinese characters. With these changes `Bytes.toString` is now faster than Haxe 4.3.7.

10k iterations:
latest: 0.09s
these: 0.07s

`Bytes.ofString` is still a fair bit slower than 4.3.7 though.

10k iterations:
latest: 0.05s
these: 0.12s

This was my benchmark code for reference.

```haxe
class Main {
    static function main() {
        // final bytes = File.getBytes("C:\\Users\\AidanLee\\Desktop\\test\\lorem.txt");
        final string = File.getContent("C:\\Users\\AidanLee\\Desktop\\test\\lorem.txt");

        Timer.measure(() -> {
            for (_ in 0...10000) {
                final _ = Bytes.ofString(string);
                // final _ = bytes.toString();
            }
        });
    }
}
```

[lorem.txt](https://github.com/user-attachments/files/25192110/lorem.txt)

tobil4sk:
> On length estimation, the one downside of this would be strings which are actually just under the "large object" size threshold of 2k bytes but are estimated to be higher. In this case those strings would go through the large object allocator rather than the thread local one.

Yeah, this is where the tradeoff lies. `__hxcpp_bytes_of_string` loops and pushes to an array, so it does constant reallocations. This patch avoids reallocations at the expense of overallocating.

> Bytes.ofString is still a fair bit slower than 4.3.7 though.

For some reason, it's the other way around for me:

ofString
| |10,000 | 1,000,000| 
|- | - | - |
| patched haxe/hxcpp | 0.06568s | 5.673s |
| haxe 4.3.7 | 0.07428s | 7.339s | 

toString
| |10,000 | 1,000,000| 
|- | - | - |
| patched haxe/hxcpp | 0.0674s | 6.107s |
| haxe 4.3.7 | 0.05319s | 4.535s |

tobil4sk:
With this version I seem to get better results than 4.3.7 for both ofString and toString:

ofString
| | 10,000 | 1,000,000| 
|- | - | - |
| patched haxe/hxcpp | 0.04984s | 6.230s |
| haxe 4.3.7 | 0.05843s | 7.230s | 



toString
| | 10,000 | 1,000,000| 
|- | - | - |
| patched haxe/hxcpp | 0.04634s | 4.068s |
| haxe 4.3.7 | 0.04579s | 4.629s |

Issue #1170, opened 2024-12-24
[Linux] Error when compiling with `HXCPP_TRACY`


NotHyper-474:
### Error
```
Error: In file included from /.../TurnBasedRPG/.haxelib/hxcpp/git/src/hx/TelemetryTracy.cpp:2:
/.../TurnBasedRPG/.haxelib/hxcpp/git/include/hx/TelemetryTracy.h:9:10: fatal error: /.../TurnBasedRPG/export/linux/obj/obj/linux64/__pch/runtime/hxcpp.h: File or directory does not exist
    9 | #include <hxcpp.h>
      |          ^~~~~~~~~
compilation terminated.
```
##
Adding `#include <hx/TelemetryTracy.h> to hxcpp like below fixes the error for me but I'm not sure if that's the correct way to solve it.
```diff
diff --git a/include/hxcpp.h b/include/hxcpp.h
index c29e6e96..2b277506 100755
--- a/include/hxcpp.h
+++ b/include/hxcpp.h
@@ -343,6 +343,7 @@ typedef PropertyAccessMode PropertyAccess;
 #include "Enum.h"
 #include <hx/Interface.h>
 #include <hx/Telemetry.h>
+#include <hx/TelemetryTracy.h>
 #if defined(__OBJC__) && defined(HXCPP_OBJC)
   #include <hx/ObjcHelpers.h>
 #endif
```

Aidan63:
This will probably be related to PCH. I've found that most hxcpp.h includes need to be guarded with `#ifndef HXCPP_H` to not run into odd issues like this (search for that in hxcpp to see examples).

PR #1312, opened 2026-03-03
Update Tracy to 0.13.1 protocol

### LINKED ISSUE #1268: Update tracy protocol to 0.13.0
HXCPP is not compatible with the latest version of Tracy, which uses the v0.13.0 protocol. This forces you to use the old version of the app.

NotHyper-474:
Closes #1268

Only tested on Arch Linux at the moment.

image


include/hxcpp.h
```diff
@@ -347,6 +347,9 @@ typedef PropertyAccessMode PropertyAccess;
 #include "Enum.h"
 #include <hx/Interface.h>
 #include <hx/Telemetry.h>
+#if defined(HX_LINUX) && defined(HXCPP_TRACY)
+#include <hx/TelemetryTracy.h>
+#endif
```
Aidan63: What's the rational for this addition? We don't want to pull the tracy header into the global header even if it's the chosen implementation.

tobil4sk: Apparently it solves #1170, but the error there seems confusing so maybe it's worth understanding why that happens and if there's a better solution. (Also maybe better to separate the fix from the tracy update)

NotHyper-474: I see. I've reverted the change.

MAJigsaw77 (approved):

PR #1319, opened 2026-03-14
Mark override functions

tobil4sk:
If c++11 is used, we need to mark override functions. For older c++ versions we just define the macro as empty.

Code that is only used for HXCPP_API_LEVEL 500 and above already requires c++11 as a minimum, so there we can just use the override keyword directly without the macro, as is already the case in some places.

To fully get rid of all warnings, we also need haxe to generate overrides where needed, i.e. https://github.com/HaxeFoundation/haxe/pull/12840.

Closes #1318 and #1334.


MAJigsaw77 (approved):

Aidan63:
Can you resolve the conflicts? I'll merge once that's done.

tobil4sk:
Should be up-to-date now!

PR #1325, opened 2026-03-22
Use a recursive mutex for the thread state change lock

Aidan63:
A user came across an issue with a dead lock involving `gThreadStateChangeLock`. Originally this was a `HxMutex` which is recursive, but when I switch everything over to stl mutexes I tried to figured out what actually needs to be recursive and got this one wrong.
I'm still not entirely sure why this needs to be recursive, but the user reported that their deadlock crash went away with the change.

### DIFF
--- a/src/hx/gc/Immix.cpp
+++ b/src/hx/gc/Immix.cpp
@@ -394,7 +394,7 @@ static int sgTimeToNextTableUpdate = 1;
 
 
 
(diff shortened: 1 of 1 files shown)


Aidan63:
Yeah, it would be nice to solve it so this can remain non recursive. Unfortunately the user never managed to create a small sample which reproduced the issue, so this is probably best for now.

PR #1327, opened 2026-03-23
[tools] Clean up apple sdk detection

tobil4sk:
Follow up to #1260. Unifies sdk detection across all apple targets and cleans up the logic a bit.

@joshtynjala would you be able to check that this still works correctly?


joshtynjala:
Yes, this is working correctly on my machine when building for both macOS and iOS.

PR #1329, opened 2026-03-27
[tools] Add HXCPP_NO_VENDORED flag for system libs

### LINKED ISSUE #1093: Allow using system sqlite?
Is there any way to link against the system sqlite instead of the bundled one?  Especially to get new features that are in the system one but not in the bundled version yet.

tobil4sk:
This allows using libraries provided by the system instead of the vendored sources. This can lead to a faster build and also gives more flexibility to have a newer version or for example in the context of packaging software for linux.

Closes #1093

### DIFF
--- a/src/hx/libs/regexp/Build.xml
+++ b/src/hx/libs/regexp/Build.xml
@@ -31,15 +31,23 @@
   <depend name="${this_dir}/Build.xml" dateOnly="true" />
   <cache value="true" asLibrary="true" />
 
-  <compilerflag value="-I${PCRE_DIR}"/>
+  <compilerflag value="-I${PCRE_DIR}" unless="HXCPP_NO_VENDORED" />
 
   <file name="${this_dir}/RegExp.cpp"/>
 </files>
 
 <target id="haxe">
   <files id="hxcpp_regexp" />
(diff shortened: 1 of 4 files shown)


Aidan63:
I'm not a massive fan of this feature. Linking directly against distro packages is an easy way to make projects that work inconsistently across distros (different distros will provide the same package with different features enabled) and brittle when distros update that package version.

If the idea is to make it easier to update dependencies then I think integrating something like vcpkg into hxcpp's build system would be a better route. In the small number of cases where linking to distro libs is useful (e.g. I know this software will only run on Ubuntu 26.04 and I'll decided when that changes) I think it's fine to require the user to manually modify build xml to their needs,

tobil4sk:
I'm not trying to advocate for making distro specific builds the default, but just to add the option so it can be used where it makes sense.

For example, if haxelib is built using hxcpp in future, it is useful to be able to build it in a way that works well as a distro package. Most distros prefer avoiding vendored dependencies in packages, so most packages provide configuration flags like `--with-system-libx`, which this is essentially equivalent to.

Outside of that, I would personally only use it for a quick local build, I agree it would not be recommended for any binary builds intended for manual download (where static linking is preferred).

PR #1330, opened 2026-03-31
Fix `MinGW` not showing traces in the console.

MAJigsaw77:
For some reason, unlike when the application is compiled with `MSVC`, prints do not appear in the console when using `MinGW` where the executable is launched. This PR fixes the issue by adjusting the logic so that it works on all `Windows` platforms except `WinRT`.


tobil4sk (approved):
There is not much context to why it was made MSVC only: 4a864bc36067eaab932cfadc42831dc1b9f42b4e

It's a bit different from regular c++ std behaviour to show console output for non-console apps, but since we do it for msvc it makes sense to do it for mingw too.

PR #1332, opened 2026-04-01
Remove `minimingw` haxelib

MAJigsaw77:
Its pretty self explanatory, this makes it so it now errors that it couldn't locate `MINGW_ROOT` instead of saying that `minimingw` is not installed instead, `minimingw` itself is pretty out of date imo.


tobil4sk (approved):
Makes sense, I think it's best to remove this because it doesn't work and hasn't been maintained for a long time. Also related to: https://github.com/HaxeFoundation/HaxeManual/pull/537

PR #1335, opened 2026-04-24
[msvc] Add WinRT ARM64 support and fix toolchain pathing/compilation issues

FixFlare:
## Summary

This PR adds support for building WinRT (UWP) applications for the ARM64 architecture using MSVC. It provides a new environment setup script and fixes directory-naming bugs in `hxcpp`.

### 1. New Setup Script: `msvc-winrt-arm64-setup.bat`

Added a dedicated setup script for WinRT ARM64. It specifically:

- Uses the `x64_arm64` cross-compiler (host x64, target arm64).
- Configures the `UWP` app platform via `vcvarsall.bat`.
- Automates detection of the ARM64 compiler components via `vswhere`.

### 2. BINDIR Routing for WinRT ARM64

Updated `BuildTool.hx` to set `BINDIR` to `WinRTArm64` when the `winrt` and `HXCPP_ARM64` defines are both present. This ensures compatibility with engine-level expectations (like Lime) and prevents WinRT ARM64 binaries from being placed in the x64 (`WinRT64`) folder.

### 3. Object Directory Naming

Fixed a bug in `msvc-toolchain.xml` where ARM64 builds would receive a redundant `arm` prefix (e.g., `armarm64`). It also ensures the `-rt` (WinRT) suffix is preserved and not overwritten by the architecture flag.

### 4. Core Runtime Hardening (Fixes ARM64 Compilation Errors)

- **`hxString.h`**: Added `explicit` to WinRT-specific string constructors (`Platform::String^`). This **fixes "ambiguous overload" errors** that occur on the ARM64 MSVC compiler during implicit conversion.
- **`StdLibs.cpp`**: Optimized `__trace` to use `HX_CSTRING` for static string literals. This **fixes compilation failures** related to string operator ambiguity on ARM64 and prevents potential runtime crashes in the UWP sandbox.

## Verification

- Verified successful cross-compilation of Haxe applications for ARM64 UWP on an x64 host.
- Verified correct directory structure generation for native libraries and object files.
- Verified manifest validation success with the resulting binaries.

### DIFF
--- a/include/hxString.h
+++ b/include/hxString.h
@@ -76,11 +76,11 @@ public:
    }
    #endif
    #if defined(HX_WINRT) && defined(__cplusplus_winrt)
-   inline String(Platform::String^ inString)
+   explicit inline String(Platform::String^ inString)
    {
       *this = String(inString->Data());
    }
-   inline String(Platform::StringReference inString)
+   explicit inline String(Platform::StringReference inString)
    {
       *this = String(inString.Data());
    }
--- a/src/hx/StdLibs.cpp
+++ b/src/hx/StdLibs.cpp
@@ -330,11 +330,11 @@ void __trace(Dynamic inObj, Dynamic info)
       fflush(stdout);
       String s;
       if (info == null()) {
-         s = String("?? ") + text + String("\n");
+         s = HX_CSTRING("?? ") + text + HX_CSTRING("\n");
       } else {
          String filename = Dynamic((info)->__Field(HX_CSTRING("fileName"), HX_PROP_DYNAMIC))->toString();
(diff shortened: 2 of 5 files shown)


MRK2009:
Wait, I have a suggestion, Can you try fix `__hxcpp_get_bin_dir()` for WinRT ARM64 in the [Lib.cpp](https://github.com/HaxeFoundation/hxcpp/blob/12f04e2894e13d22323eec0692a6ef59d4338761/src/hx/Lib.cpp#L393) file?
Example of the fix:
  ```
  #ifdef HXCPP_ARM64
    HX_CSTRING("WinRTArm64");
  #elif defined(HXCPP_M64)
    HX_CSTRING("WinRT64");
  #else
    HX_CSTRING("WinRT");
  #endif
  ```
  Although the emulation layer works fine, unless Haxe will ever supports Windows on ARM.

PR #1336, opened 2026-04-28
Typed thread and tls api

Aidan63:
See https://github.com/HaxeFoundation/haxe/pull/12885 for details


toolchain/haxe-target.xml
```diff
@@ -166,6 +170,9 @@
   <depend name="${HXCPP}/src/hx/Hash.h"/>
   <depend name="${HXCPP}/src/hx/Unicase.h"/>
   <depend name="${HXCPP}/src/hx/gc/GcRegCapture.h"/>
+  <depend name="${HXCPP}/src/hx/thread/ThreadImpl.hpp"/>
+  <compilerflag value="-D_CRT_SECURE_NO_DEPRECATE"/>
```
tobil4sk, line 174: This is a microsoft CRT specific flag so it should be in the msvc toolchain instead of here

Issue #1289, opened 2026-01-07
Random crashes in mac arm64 "haxe" tests


tobil4sk:
[Test MacOS (Arm64) (4.3.4) / haxe](https://github.com/HaxeFoundation/hxcpp/actions/runs/20779134226/job/59672101581?pr=1285#step:5:11)

```
/Users/runner/work/_temp/72bb127a-6426-4be2-86ed-fce9d29fd2ee.sh: line 1: 15185 Segmentation fault: 11  bin/TestMain
Error: Process completed with exit code 139.
```

[Test MacOS (Arm64) (latest) / haxe](https://github.com/HaxeFoundation/hxcpp/actions/runs/20757730218/job/59604240179#step:5:11)

```
Memory exhausted.
/Users/runner/work/_temp/d85ca6f0-43ca-4142-8d9d-60a39b2041fe.sh: line 1:  5788 Trace/BPT trap: 5       bin/TestMain
 try HXCPP_GC_BIG_BLOCKS.
Error: Process completed with exit code 133.
```

Aidan63:
I can re-create this locally in WSL, it always occurs in the TestGCThreaded test. I'm assuming what's happening is that some objects aren't being collected causing it to run out of memory in this high allocation test? I'll try and bisect which commit this started occurring on.

Aidan63:
I've gone all the way back to 05/06/2025 when the millisecond resolution timer was added and the illegal instruction error still occurs, so it seems like this has been around for a while but has only just started cropping up in the CI?

Disabling the generation GC also makes no difference, the error still occurs. Maybe we do need to enable big blocks, but I can't see why the memory would be exhausted with those tests.

tobil4sk:
Is there anything specific that needs to be done to reproduce it? or just running it a couple times until it crashes?

Aidan63:
Yeah, I've been checking out hxcpp commits and then close / relevant haxe commits and building them together (I'm compiling the hxcpp tests with debug as well).

I've then been running `while ./bin/TestMain-debug; do :; done` to run the tests until they return a non zero exit.

When attaching a debugger the error occurs when allocating one of the wrapper classes in the TestGCThreaded test, it's hitting the final path where all allocation attempts return null.

PR #1339, opened 2026-05-07
Fix Immix working memory overflow with big blocks

Tutez64:
## Summary

Fixes an overflow in Immix working memory accounting when `HXCPP_GC_BIG_BLOCKS` is enabled.

The problematic code computes the regular Immix heap size through an `int`:

```cpp
int blockSize = mAllBlocks.size() << IMMIX_BLOCK_BITS;
```
With big blocks, this can overflow once regular heap usage grows past roughly 2 GiB. The fix uses `GetWorkingMemory()`, which performs the shift through `size_t`.
This also stores related GC memory tuning thresholds as `size_t` and parses large environment variable values with `strtoull.`
## Context
In a large OpenFL/hxcpp game, the regular heap consistently failed once it passed 2 GiB with:
```
Memory exhausted.
try HXCPP_GC_BIG_BLOCKS.
```
After enabling `HXCPP_GC_BIG_BLOCKS,` the game no longer crashed immediately, but it froze for a very long time instead, making it unplayable.
With this fix, the same game can continue past that threshold. I was able to reach 12+ GiB of regular heap usage without hitting the previous stall/failure mode.
## Repro
[HxcppBigBlocksHeapOverflowRepro.zip](https://github.com/user-attachments/files/27495862/HxcppBigBlocksHeapOverflowRepro.zip)
It retains many regular hxcpp objects until multi-GiB heap usage.
The repro is not a perfect match for the original game: without the fix, it usually stalls around 7-8 GiB rather than just past 2 GiB. However, it still demonstrates the same failure mode and shows that the fix lets the program continue to the configured target.

### DIFF
--- a/include/hx/GC.h
+++ b/include/hx/GC.h
@@ -158,10 +158,10 @@ extern int gPauseForCollect;
 
 
 // Minimum total memory - used + buffer for new objects
-extern int sgMinimumWorkingMemory;
+extern size_t sgMinimumWorkingMemory;
 
 // Minimum free memory - not counting used memory
-extern int sgMinimumFreeSpace;
+extern size_t sgMinimumFreeSpace;
 
 // Also ensure that the free memory is larger than this amount of used memory
 extern int sgTargetFreeSpacePercentage;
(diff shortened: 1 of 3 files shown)


tobil4sk:
@Aidan63 I wonder if this could be relevant to our memory exhaustion in the tests? #1289

Aidan63:
Maybe, but this seems to be about big blocks specifically. Seems like a good idea to run those that flaky haxe test through clang with all the sanitisers enabled to see if any over / under flow is triggered though.

PR #1342, opened 2026-05-10
Add support to cross-compile `Windows` via `MinGW` from `MacOS`

MAJigsaw77:
Its pretty self explenetory, this makes it so `MinGW` can now be compiled from `MacOS` and not only from `Linux` or `Windows`, I dont why this hasn't been done earlier


tobil4sk (approved):
@Aidan63 Would it be possible to merge this? I would like to avoid merge conflicts with #1361.

PR #1343, opened 2026-05-11
iphonesim HXCPP_ARM64: distinct LIBEXTRA/objdir, arm64 dll linker (follow-up to 3668203c)

AxGord:
## Summary

**Rescoped after 3668203c** — the core arm64 iOS simulator support
(`-arch arm64` + `-mios-simulator-version-min`) landed upstream in
3668203c, which supersedes the compiler-flag part of this PR. Upstream
master has been merged in and this branch's equivalent `-target` triple
sections were dropped in favor of upstream's approach.

What remains (8 insertions, 5 deletions across 2 files):

- **Distinct `LIBEXTRA` for arm64-sim** (`finish-setup.xml`): with
  `HXCPP_ARM64` set, simulator libs currently get the plain `.iphonesim`
  suffix — colliding with the i386 name. This PR names them
  `.iphonesim-arm64` so arm64/x86_64/i386 simulator binaries can coexist
  (matters for multi-slice workflows and for Lime's NDLL naming).
- **Per-arch objdir** (`iphonesim-toolchain.xml`): all simulator archs
  currently share `obj/iphonesim/`; switching arch without a clean
  rebuild mixes object files. arm64-sim gets `obj/iphonesim…-arm64/`.
- **dll linker arm64** (`iphonesim-toolchain.xml`): the `dll` linker
  still only emits `-arch i386`/`-arch x86_64` — with `HXCPP_ARM64` it
  produces an i386 dylib. Rarely hit (sim builds are `static_link` by
  default), but wrong; this adds the `arm64` case with the same
  `if`/`unless` guards as the compiler.

With `HXCPP_ARM64` unset, everything is byte-identical to master.

We've been carrying arm64-sim support in our `soccertutor/hxcpp` fork for
~1.5 years, used in production OpenFL+AIR apps that ship to iPad; built
and tested on Apple Silicon (M1 Pro) against Xcode 15/16 SDKs, iOS
simulator runtimes 16/17/18. `xmllint --noout` confirms both files are
well-formed.

## Use

```bash
haxelib run lime build ios -simulator -Dhxcpp_arm64
# or direct hxcpp:
haxe -cp <path> -cpp out -DHXCPP_ARM64 -Diphonesim
```

## Context for follow-up

This is the foundation for an Apple Silicon support PR queued for
`openfl/lime` (NDLL fallback chain, auto-OpenAL backend selection),
which keys off the arm64-sim `LIBEXTRA` naming.

<details>
<summary>Original PR description (pre-3668203c)</summary>

The original scope added full arm64 simulator support via
`-target arm64-apple-ios${version}-simulator` compiler flags, separate
`LIBEXTRA`/objdir, and the dll linker arch. The compiler-flag half was
implemented upstream independently in 3668203c using
`-arch arm64` + `-mios-simulator-version-min`, which achieves the same
correct `LC_BUILD_VERSION` simulator platform marking.

</details>


toolchain/iphonesim-toolchain.xml
```diff
@@ -38,16 +38,24 @@
...
+    <flag value="-isysroot"/>
+    <flag value="${DEVELOPER_DIR}/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator${IPHONE_VER}.sdk"/>
+  </section>
+  <!-- x86_64/i386 simulator: original flags -->
+  <section unless="HXCPP_ARM64">
+    <flag value="-arch"/>
+    <flag value="i386" unless="HXCPP_M64"/>
+    <flag value="x86_64" if="HXCPP_M64"/>
+    <flag value="-isysroot"/>
+    <flag value="${DEVELOPER_DIR}/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator${IPHONE_VER}.sdk"/>
```
tobil4sk: Could the `-isysroot` part be extracted outside the sections since it seems to be common to both?

MRK2009 (approved):
Now it's much better, I think it would be cool if ARM64 support for iOS simulator going to be implemented in lime.

toolchain/iphonesim-toolchain.xml
```diff
@@ -69,16 +75,18 @@
 
   <outflag value="-o"/>
   <ext value=".o"/>
-  <objdir value="obj/iphonesim${OBJGCC}${OBJDBG}/" />
+  <objdir value="obj/iphonesim${OBJGCC}${OBJDBG}-64/" if="HXCPP_ARM64" />
```
tobil4sk: Here we want the suffix to differentiate x86_64 from arm64. Maybe it's better to use something like `-arm64`, since they are both 64 bit architectures so `-64` isn't very descriptive.

AxGord: Good point — renamed to `-arm64` in 06609048. It now also matches the `.iphonesim-arm64` LIBEXTRA suffix in finish-setup.xml.

MRK2009:
This commit causes the pr have merge conflicts
https://github.com/HaxeFoundation/hxcpp/commit/3668203c4dfac631194201d8f426a0bbeb81e33d

AxGord:
Conflicts resolved — merged master in (fc373fea). Since 3668203c implements the arm64 simulator compiler flags upstream (thanks!), I dropped this branch's equivalent `-target` sections and rescoped the PR to the remaining gaps: distinct `.iphonesim-arm64` LIBEXTRA, per-arch sim objdir, and the `arm64` case in the `dll` linker (which otherwise still emits `-arch i386` under `HXCPP_ARM64`). Diff vs master is now +8/−5 across the same 2 files; PR description updated accordingly.

PR #1349, opened 2026-06-22
Fixes for OpenBSD build

singpolyma:


HomuHomu833:
ummm is hxcpp meant to run in BSD excluding macOS?

tobil4sk:
I think currently there isn't full support for BSD operating systems in the build tool but it seems like something that wouldn't be too difficult to have with the proper toolchain files.

singpolyma:
My stuff works fine on OpenBSD with these changes

PR #1160, opened 2024-10-24
Remove `_CRT_SECURE_NO_DEPRECATE` from the `haxe` toolchain

MAJigsaw77:

PR #1351, opened 2026-06-25
Use fallback ThreadImpl for mingw

tobil4sk:
Possibly a regression from: #1336.

Mingw is has a pthread handle as `std::thread::native_handle_type` so it cannot be used in APIs that expect `HANDLE`:
```
Error: /.../hxcpp/src/hx/thread/ThreadImpl.win32.cpp: In member function ‘virtual String hx::thread::ThreadImpl_obj::getName()’:
/.../hxcpp/src/hx/thread/ThreadImpl.win32.cpp:14:52: error: invalid conversion from ‘std::thread::native_handle_type’ {aka ‘long long unsigned int’} to ‘HANDLE’ {aka ‘void*’} [-fpermissive]
   14 |         auto result = GetThreadDescription(native->handle, &buffer);
      |                                            ~~~~~~~~^~~~~~
      |                                                    |
      |                                                    std::thread::native_handle_type {aka long long unsigned int}
In file included from /usr/x86_64-w64-mingw32/include/winbase.h:29,
                 from /usr/x86_64-w64-mingw32/include/windows.h:70,
                 from /.../hxcpp/src/hx/thread/ThreadImpl.win32.cpp:2:
/usr/x86_64-w64-mingw32/include/processthreadsapi.h:367:58: note: initializing argument 1 of ‘HRESULT GetThreadDescription(HANDLE, WCHAR**)’
  367 |   WINBASEAPI HRESULT WINAPI GetThreadDescription (HANDLE hThread, PWSTR *ppszThreadDescription);
      |                                                   ~~~~~~~^~~~~~~
/.../hxcpp/src/hx/thread/ThreadImpl.win32.cpp: In member function ‘virtual void hx::thread::ThreadImpl_obj::setName(const String&)’:
/.../hxcpp/src/hx/thread/ThreadImpl.win32.cpp:39:52: error: invalid conversion from ‘std::thread::native_handle_type’ {aka ‘long long unsigned int’} to ‘HANDLE’ {aka ‘void*’} [-fpermissive]
   39 |         auto result = SetThreadDescription(native->handle, name.wchar_str(&buffer));
      |                                            ~~~~~~~~^~~~~~
      |                                                    |
      |                                                    std::thread::native_handle_type {aka long long unsigned int}
/usr/x86_64-w64-mingw32/include/processthreadsapi.h:366:58: note: initializing argument 1 of ‘HRESULT SetThreadDescription(HANDLE, PCWSTR)’
  366 |   WINBASEAPI HRESULT WINAPI SetThreadDescription (HANDLE hThread, PCWSTR lpThreadDescription);
      |                                                   ~~~~~~~^~~~~~~
/.../hxcpp/src/hx/thread/ThreadImpl.win32.cpp: In constructor ‘hx::thread::ThreadImpl_obj::Native::Native()’:
/.../hxcpp/src/hx/thread/ThreadImpl.win32.cpp:55:81: error: invalid conversion from ‘HANDLE’ {aka ‘void*’} to ‘std::thread::native_handle_type’ {aka ‘long long unsigned int’} [-fpermissive]
   55 | hx::thread::ThreadImpl_obj::Native::Native() : thread(), handle(GetCurrentThread()) {}
      |                                                                 ~~~~~~~~~~~~~~~~^~
      |                                                                                 |
      |                                                                                 HANDLE {aka void*}
```

### DIFF
--- a/toolchain/haxe-target-threads.xml
+++ b/toolchain/haxe-target-threads.xml
@@ -9,7 +9,8 @@
 
   <section if="windows">
     <file name="src/hx/thread/CountingSemaphore.win32.cpp"/>
-    <file name="src/hx/thread/ThreadImpl.win32.cpp"/>
+    <file name="src/hx/thread/ThreadImpl.win32.cpp" unless="mingw" />
+    <file name="src/hx/thread/ThreadImpl.noop.cpp" if="mingw" />
   </section>
 
   <section if="apple">
--- a/toolchain/haxe-target.xml
+++ b/toolchain/haxe-target.xml
@@ -172,7 +172,6 @@
   <depend name="${HXCPP}/src/hx/Unicase.h"/>
   <depend name="${HXCPP}/src/hx/gc/GcRegCapture.h"/>
   <depend name="${HXCPP}/src/hx/thread/ThreadImpl.hpp"/>
-  <compilerflag value="-D_CRT_SECURE_NO_DEPRECATE"/>
 
   <compilerflag value="-DHX_UNDEFINE_H" />
 


Aidan63:
Could we use c++ casts instead of the old c style ones? Also, is there some sort of mingw define to put these changes behind? The msvc stl returns a `HANDLE` for the native thread type so it don't need any casting.

tobil4sk:
> The msvc stl returns a HANDLE for the native thread type so it don't need any casting.

Hm, that makes me think maybe I'll need to look into this more properly to make sure this fix is correct. Perhaps the actual problem is that the native handle on mingw is a pthread handle, in which case the cast is not a sufficient fix.

Aidan63:
From a quick google it seems mingw provides two threading implementations, a posix one (the default) and a win32 one, but only the posix one provides a C++11 threading implementation.

https://github.com/boostorg/fiber/issues/172

There are some third party projects which provides C++11 threading support for the win32 implementation though. https://github.com/meganz/mingw-std-threads

So it does seem like this change just silences the warning but it won't actually work as it's just casting a pthread handle to a win32 HANDLE. This has done nothing to change my opinion that mingw and it's ilk are a massive nightmare and not worth it.

tobil4sk:
I made it fall back to the noop implementation on mingw which seems like the easiest fix, given that this feature is not fully supported anyway.

I've also removed the `-D_CRT_SECURE_NO_DEPRECATE` flag from haxe-target.xml, since it was being added globally but is only relevant to windows msvc builds and it is already added by `msvc-toolchain.xml`. (It was already removed in #1160, but perhaps reintroduced in #1336 due to a merging accident.)

tobil4sk:
Thanks!

PR #1355, opened 2026-07-12
Support for separate Linux ARM architectures

MRK2009:
Redo of Lily's pr https://github.com/HaxeFoundation/hxcpp/pull/1159

You can finally use the arm architecture flags to able to compile on linux arm instead of only running it on different architectures to able to compile.

It seems lime has some armv7 linux support, but not that much because it's still supports arm64.

### DIFF
--- a/tools/hxcpp/BuildTool.hx
+++ b/tools/hxcpp/BuildTool.hx
@@ -66,6 +66,7 @@ class BuildTool
    var mNvccLinkFlags:Array<String>;
    var mDirtyList:Array<String>;
    var arm64:Bool;
+   var armv7:Bool;
    var m64:Bool;
    var m32:Bool;
    var defaultCStandard:Null<Int>;
@@ -138,8 +139,9 @@ class BuildTool
       m64 = mDefines.exists("HXCPP_M64");
       m32 = mDefines.exists("HXCPP_M32") || mDefines.exists("HXCPP_X86");
       arm64 = mDefines.exists("HXCPP_ARM64");
-      var otherArmArchitecture = mDefines.exists("HXCPP_ARMV6") || mDefines.exists("HXCPP_ARMV7") || mDefines.exists("HXCPP_ARMV7S");
-      if (m64==m32 && !arm64 && !otherArmArchitecture)
+      armv7 = mDefines.exists("HXCPP_ARMV7");
+      var otherArmArchitecture = mDefines.exists("HXCPP_ARMV6") || mDefines.exists("HXCPP_ARMV7S");
+      if (m64==m32 && !arm64 && !armv7 && !otherArmArchitecture)
       {
          var arch = mDefines.get("HXCPP_ARCH");
          if (arch!=null)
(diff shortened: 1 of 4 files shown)


MRK2009:
I think it's necessary enough to be merged. 🤔

MRK2009:
Sorry for committing too much, I was trying to think of an alternative way to make it work.

docs/build_xml/Defines.md
```diff
@@ -66,8 +66,6 @@ Defines affecting target architecture.
 | *HXCPP_ARM64*           | Compile arm-based devices for 64 bits |
 | *HXCPP_ARMV7*           | Compile arm-based devices for armv7 |
 | *HXCPP_ARMV7S*          | Compile arm-based devices for armv7s |
-| *HXCPP_LINUX_ARMV7*     | Run on a linux ARMv7 device |
-| *HXCPP_LINUX_ARM64*     | Run on a linux ARM64 device |
```
tobil4sk, line 70: Maybe we should instead put these in a new deprecated/removed define section, people might be using this flag and it would be confusing if they search it up in the documentation and there is no trace of it.

tools/hxcpp/BuildTool.hx
```diff
@@ -2138,21 +2142,21 @@ class BuildTool
             defines.set("toolchain","linux");
             defines.set("linux","linux");
 
-            if (defines.exists("HXCPP_LINUX_ARMV7"))
```
tobil4sk, line 2142: It would be ok to remove this for hxcpp 5, but until then it would be best to avoid the breaking change and to continue to respect this define.

MRK2009: I'll look into it.

MRK2009:
I brought back the defines.
So it won't break.

tobil4sk (approved):

toolchain/common-defines.xml
```diff
@@ -5,7 +5,7 @@
  <flag value="-DHXCPP_STATIC_CFFI" if="HXCPP_STATIC_CFFI" tag="cffi" />
  <flag value="-DHXCPP_ARM64" if="HXCPP_ARM64"/>
  <flag value="-DHXCPP_ARMV7" if="HXCPP_ARMV7"/>
- <flag value="-DHXCPP_M64" if="HXCPP_M64||HXCPP_ARM64||HXCPP_X86_64"/>
+ <flag value="-DHXCPP_M64" if="HXCPP_M64||HXCPP_X86_64" unless="HXCPP_ARM64" />
```
tobil4sk: Why is this needed? Currently `HXCPP_M64` should be a general 64-bit define

MRK2009: Oh sorry, I thought it would not work, but it seems the previous one works fine, so I revert the change I made.

MRK2009: Maybe I'll test it properly later.

MRK2009:
Yeah, it looks fine after I tested it properly(kinda).

PR #1356, opened 2026-07-17
Fix a macOS latest build from zlib be outdated

Paopun20:
just update zlib for fix macos build

PR #1358, opened 2026-07-26
Update changelog

Aidan63:
I've gone through the git log since the last time I updated the changelog and added all the missing stuff. In a day or two I'll merge this in and then take the resulting CI artefact and use that as a basis for a new haxelib release.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -11,10 +11,36 @@
 * Added support for Android NDKs higher than 21
 * Added x86_64 support to older Android NDKs
 * Added optional detaching of main thread
+* Added flushes after each trace and println call
+* Added HXCPP_ANDROID_PLATFORM define which deprecated PLATFORM_NUMBER
+* Added support for ARMv7 ABI in Android GCC NDKs
+* Added HXCPP_NO_IMPLIB to avoid generation of .exp and .lib files on Windows
+* Added hiding CFFI symbols when static linking
+* Added callback function for cppia script loading
+* Added new log level, setup, which is between info and verbose
+* Added custom stack trace support
+* Added millisecond resolution timer
+* Added AtomicObject implementation
+* Added 16KB as the max page size on Android
+* Added HXCPP_ANDROID_NO_16K_PAGES to not use 16KB as the max page size
+* Added allocation alignment with HXCPP_ALIGN_ALLOC by default
+* Added typed zip API
+* Added check for invalid ANDROID_NDK_ROOT
+* Added string encodings API
+* Added Haxe friendly rooting API
+* Added PCH support for clang
+* Added HXCPP_CXX_STANDARD define
+* Added cpp_standard and c_standard xml attributes
+* Added typed semaphore API
+* Added override attributes to relevant functions
+* Added HXCPP_GENERATE_MSVC to generate a Visual Studio solution
+* Added String::fromCharCode to encode UTF16 surrogates unchanged
+* Added typed thread and thread local storage API
 
 * Updated mbedtls to 2.28.2
-* Updated sqlite to 3.40.1
-* Updated zlib to 1.2.13
+* Updated sqlite to 3.51.2
+* Updated zlib to 1.3.1
+* Updated tracey to 0.13.1
 
 * Fixed SSL socket non blocking handshake throwing an exception on 64bit Windows
 * Fixed Windows 64bit architecture detection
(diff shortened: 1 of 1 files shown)


Aidan63:
The failing CI has reminded me I want to try and track that down that crash before the new release since it only seems to occur on 4.3.7.

Changes.md
```diff
@@ -32,6 +58,33 @@
...
+* Fixed UTF8 output in Windows consoles
+* Fixed HXCPP_M64 being set if you defined HXCPP_X86
+* Fixed base types not being fully qualified in macros
+* Fixed cppia bindir for ARM64
+* Fixed fastcall attribute being added on Linux aarch64
+* Fixed Macsox26 SDK not being located
+* Fixed blocking SQLite calls not being made in a GC free zone
+* Fixed MinGW toolchain RC file handling
 
 * Removed Haxe 3 support
```
tobil4sk, line 89: Haxe 3.4.7 should still build fine, it's Haxe 3.3 support specifically that was removed in #1083 (and was probably already broken before).

Aidan63: Does it actually work? My intent was to remove all Haxe 3 code paths. I'm happy to still have it say removed all Haxe 3 to stop someone trying to use it.

tobil4sk: Yes, Haxe 3.4.7 builds with nightly hxcpp and passes all Haxe unit tests, with the exception of this test that ended up being removed because it was incorrect:
https://github.com/HaxeFoundation/haxe/blob/bb7b827a9c135fbfd066da94109a728351b87b92/tests/unit/src/unit/TestReflect.hx#L114

https://github.com/HaxeFoundation/haxe/commit/6b0fa1005dbb3fd3f027f8ea06a12007537b51e9

I still have to sometimes compare behaviour with Haxe 3.4.7, and it is useful to be able to do that without having to dig out an old version of hxcpp. Lime also still supports Haxe 3.4.7 (until its next major release) and relies on being able to use modern hxcpp for it. Hxcpp's backwards compatibility is something that makes it a stable and reliable target.

> I'm happy to still have it say removed all Haxe 3 to stop someone trying to use it.

I think the tool emitting a warning like "Haxe 3.4 support is deprecated and will be removed in a future hxcpp release" would be a more effective measure to encourage users to migrate than an inaccurate changelog entry.

Changes.md
```diff
@@ -11,10 +11,36 @@
...
+* Added check for invalid ANDROID_NDK_ROOT
+* Added string encodings API
+* Added Haxe friendly rooting API
+* Added PCH support for clang
+* Added HXCPP_CXX_STANDARD define
+* Added cpp_standard and c_standard xml attributes
+* Added typed semaphore API
+* Added override attributes to relevant functions
+* Added HXCPP_GENERATE_MSVC to generate a Visual Studio solution
+* Added String::fromCharCode to encode UTF16 surrogates unchanged
```
tobil4sk, line 37: This one is more of a bug fix than an addition

Changes.md
```diff
@@ -11,10 +11,36 @@
...
+* Added typed thread and thread local storage API
 
 * Updated mbedtls to 2.28.2
-* Updated sqlite to 3.40.1
-* Updated zlib to 1.2.13
+* Updated sqlite to 3.51.2
+* Updated zlib to 1.3.1
+* Updated tracey to 0.13.1
```
tobil4sk, line 43: Tracey is a new feature, so I think it makes more sense to say the version in the "Added" entry than to say it was updated

Changes.md
```diff
@@ -32,6 +58,33 @@
...
+* Fixed out of bounds access in String::fromCharCode
+* Fixed .so ext not being added on Android clang toolchain
+* Fixed *= in cppia jit with boxed numbers
+* Fixed HXCPP_CPP11 not being respected on Android toolchains
+* Fixed integer overflow arithmetic assignments in cppia
+* Fixed cppia overridden scriptable functions
+* Fixed Pointer.ofArray with empty array
+* Fixed semaphores and condition variables not having GC free zones around blocking calls
+* Fixed HXCPP_DEBUG_LINK and HXCPP_OPTIMIZE_LINK not being respected on Linux
+* Fixed UTF8 output in Windows consoles
```
tobil4sk, line 80:
```suggestion
* Fixed unicode output in Windows consoles
```
UTF8 isn't really involved at all here, the unicode strings are currently utf-16 encoded and go straight to the windows console without conversion: #1307

Changes.md
tobil4sk, line 47: "critial" -> "critical"

tobil4sk (commented):
In general, I wonder the best way to handle the Haxe 5 api in this next hxcpp release. Really it is experimental/unstable support for Haxe 5, and will likely break again in future nightly haxe before the final haxe 5 release. Maybe it is worth having a separate part of the changelog for haxe 5 related additions, mentioning that they are still subject to change.

Also, I know this hasn't been done previously, but since some of the entries are reworded from the original commit messages, it would be useful to have links to the commit/PR for reference. I can help with this if needed.

MRK2009:
Also zlib needs to be updated too by merging this pr(https://github.com/HaxeFoundation/hxcpp/pull/1356) and then the change version of zlib from 1.3.1 to 1.3.2 in the changelog.

PR #1359, opened 2026-07-27
Add HXCPP_CATCH_SEGV to test hxml

Aidan63:
Seeing if we can get more info on #1289


Aidan63:
No extra info, I'll just try running the tests on Linux until they fail.

PR #1361, opened 2026-08-02
Add support for LLVM-based MSVC builds

tobil4sk:
This adds a `-D HXCPP_MSVC_LLVM` flag which allows creating msvc builds using `clang-cl` instead of `cl`.

This is useful for native windows builds, however, it also allows cross-compiling msvc binaries from linux by passing `-D HXCPP_MSVC_ROOT=/path/to/msvc-root` (msvc libs/headers can be downloaded using `xwin` for example).

### DIFF
--- a/tools/hxcpp/Compiler.hx
+++ b/tools/hxcpp/Compiler.hx
@@ -697,7 +697,7 @@ class Compiler
       mPCH = inPCH;
       createCompilerVersion();
       final regex = ~/clang/i;
-      if (inPCH == null && regex.match(mCompilerVersionString)) {
+      if (inPCH != null && regex.match(mCompilerVersionString) && !BuildTool.isMsvcLlvm()) {
          mPCH = "clang";
       }
       switch (mPCH) {
--- a/tools/hxcpp/Compiler.hx
+++ b/tools/hxcpp/Compiler.hx
@@ -695,7 +695,7 @@ class Compiler
       mPCH = inPCH;
       createCompilerVersion();
       final regex = ~/clang/i;
-      if (inPCH != null && regex.match(mCompilerVersionString)) {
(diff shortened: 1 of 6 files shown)


tools/hxcpp/Compiler.hx
```diff
@@ -695,7 +697,7 @@ class Compiler
       mPCH = inPCH;
       createCompilerVersion();
       final regex = ~/clang/i;
-      if (inPCH != null && regex.match(mCompilerVersionString)) {
+      if (inPCH == null && regex.match(mCompilerVersionString)) {
```
tobil4sk: @Apprentice-Alchemist The previous version came from your PR #1299. It meant that even if msvc-toolchain specifies msvc style pch handling, it would be ignored and the `clang` in the version output would take priority for clang-cl. Was there a reason why the compiler version output to take priority, or was it just a typo?

Apprentice-Alchemist: Apple and Android toolchains set `USE_PRECOMPILED_HEADERS` to gcc (by including `gcc-toolchain.xml`) so that needs to be overriden for precompiled headers to work.
Now those specific cases could be fixed in the toolchain, but I wouldn't be surprised if there are other systems that alias gcc/g++ to Clang, so compiler version output should probably take priority.

tobil4sk: I've changed it back so it preserves the old behaviour for android and mac. It seems like something that should be cleaned up in future though. Thanks!

MTUnixic:
hope this gets merged

HomuHomu833:
could u guys prioritize [llvm-mingw](https://github.com/mstorsjo/llvm-mingw) too? it would be great.

MRK2009:
It got conflicted.

PR #1366, opened 2026-08-09
Fix cppia JIT "Bad move target" on untyped register moves

MeguminBOT:
Any binary subtraction inside a ternary that is converted to String or object triggers it. 
Multiplication, division, unary minus and addition are unaffected, as is hoisting the subtraction into a local first.

`CppiaCompiler::convert` moved between two registers without giving either a type in three places. 
`move()` decides what to emit from `getCommonType()`, and `getCommonType(jtAny, jtAny)` returns `jtAny`, which the
switch answers with `setError("Bad move target")`.

All three sites are guarded by `inSrc.uses(SLJIT_R1)` or `inSrc==sJitTemp1`, so they are only reached when the source value happens to live in R1.
`OpAdd::genCode` leaves its int result in `sJitTemp0` and misses the guard, while `OpSub::genCode` uses `sJitTemp1` and hits it, which is why addition compiles and subtraction does not.

### Reproducable steps
```haxe
class Repro {
    static function f(a:Int, b:Int, flag:Bool):String {
        return "x" + (flag ? 0 : a - b);
    }
    public static function main() Sys.println(f(5, 3, false));
}
```

### Run
```
haxe -main Repro -cppia repro.cppia
haxe -main cpp.cppia.Host -D scriptable -dce no -cpp host
host/Host.exe -jit repro.cppia    Error : Bad move target
host/Host.exe      repro.cppia    x2
```

### DIFF
--- a/src/hx/cppia/CppiaCompiler.cpp
+++ b/src/hx/cppia/CppiaCompiler.cpp
@@ -898,7 +898,7 @@ public:
                {
                   if (inSrc.uses(SLJIT_R1))
                   {
-                     move(sJitArg0, inSrc);
+                     move(sJitArg0.as(jtInt), inSrc.as(jtInt));
                      add( sJitTemp1, inTarget.getReg(), inTarget.offset );
                      callNative( (void *)intToStr, sJitArg0.as(jtInt), sJitTemp1.as(jtPointer));
                   }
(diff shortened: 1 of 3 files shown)


tobil4sk:
It's fine if you just force push your new commits here, without closing and opening an new PR

tobil4sk:
They should also run on push, but they might require approval for first time contributors

PR #1367, opened 2026-08-13
Give a cppia member field declared Bool its boolean storage

MeguminBOT:
### The problem

A class member declared `Bool` in a cppia module gets an integer slot, so reading it back through
reflection gives you `1` instead of `true`. A static field of the same declared type, in the same
module, gives you `true`:

```
member 1, static true
```

Conditions and comparisons still work, since they are happy with the integer. Anything that looks at
the value itself does not.

### Why

`CppiaVar::linkVarTypes` has two forms:

* The **static** form calls `fieldStorageFromType`, which has a case for `Bool` and returns `fsBool`.
* The **member** form switches on `exprType` instead. By the time it runs, `TypeData::link` has
  already mapped the `Bool` type name to `etInt`, so the field ends up as `fsInt`.

That is why the two fields above disagree.

### The fix

In `src/hx/cppia/CppiaVars.cpp`, the member form now calls `fieldStorageFromType` too:

```cpp
storeType = typeId==0 ? fsObject : fieldStorageFromType(type);

ioOffset += sTypeSize[exprType];
```

The slot is still int sized, so the layout and the `AlignOffset` call above it do not change.

### Test

`test/cppia` covers it. `ClientBoolField` in `Client.hx` holds a `Bool` member and a `Bool` static,
and `testBoolMemberStorage` in `cases/TestCommon.hx` reads both back through `Reflect.field`.

```
cd test/cppia
haxe compile-host.hxml
haxe compile-client.hxml
cd bin && ./CppiaHost.exe client.cppia
```

Without the fix that test fails with `Member Bool did not read back as a boolean`, and the static
assertion beside it still passes, which is the pair that shows the two forms disagreeing. Same result
with or without `-jit`.

### Reproducing by hand

`Script.hx`, built with `haxe -m Script --cppia script.cppia`:

```haxe
class Holder {
   public var flag:Bool = true;
   public static var staticFlag:Bool = true;
   public function new() {}
}

class Script {
   public static function run():String {
      return 'member ' + Std.string(Reflect.field(new Holder(), 'flag'))
         + ', static ' + Std.string(Reflect.field(Holder, 'staticFlag'));
   }

   public static function main():Void {}
}
```

Load it from a host built with `-D scriptable`, using
`cpp.cppia.Module.fromData(bytes).boot()`, and call `Script.run` through reflection.

| | result |
| --- | --- |
| before | `member 1, static true` |
| after | `member true, static true` |

Same result with the JIT on or off, since the storage is decided at link time.

**Let me know if I got anything wrong!**

### DIFF
--- a/src/hx/cppia/CppiaVars.cpp
+++ b/src/hx/cppia/CppiaVars.cpp
@@ -180,14 +180,17 @@ void CppiaVar::linkVarTypes(CppiaModule &cppia, int &ioOffset)
       AlignOffset(exprType, ioOffset);
       offset = ioOffset;
       
-      switch(exprType)
+      storeType = typeId==0 ? fsObject : fieldStorageFromType(type);
+
+      switch(storeType)
       {
-         case etInt: ioOffset += sizeof(int); storeType=fsInt; break;
-         case etFloat: ioOffset += sizeof(Float);storeType=fsFloat;  break;
-         case etString: ioOffset += sizeof(String);storeType=fsString;  break;
-         case etObject: ioOffset += sizeof(hx::Object *);storeType=fsObject;  break;
-         case etVoid:
-         case etNull:
+         case fsBool: ioOffset += sizeof(int); break;
+         case fsByte: ioOffset += sizeof(int); break;
+         case fsInt: ioOffset += sizeof(int); break;
+         case fsFloat: ioOffset += sizeof(Float); break;
+         case fsString: ioOffset += sizeof(String); break;
+         case fsObject: ioOffset += sizeof(hx::Object *); break;
+         case fsUnknown:
             break;
       }
    }
--- a/src/hx/cppia/CppiaVars.cpp
+++ b/src/hx/cppia/CppiaVars.cpp
@@ -182,17 +182,7 @@ void CppiaVar::linkVarTypes(CppiaModule &cppia, int &ioOffset)
       
       storeType = typeId==0 ? fsObject : fieldStorageFromType(type);
 
-      switch(storeType)
-      {
-         case fsBool: ioOffset += sizeof(int); break;
-         case fsByte: ioOffset += sizeof(int); break;
-         case fsInt: ioOffset += sizeof(int); break;
-         case fsFloat: ioOffset += sizeof(Float); break;
-         case fsString: ioOffset += sizeof(String); break;
-         case fsObject: ioOffset += sizeof(hx::Object *); break;
-         case fsUnknown:
-            break;
-      }
+      ioOffset += sTypeSize[exprType];
    }
 }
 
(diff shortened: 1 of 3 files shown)


src/hx/cppia/CppiaVars.cpp
```diff
@@ -180,14 +180,17 @@ void CppiaVar::linkVarTypes(CppiaModule &cppia, int &ioOffset)
...
-         case etObject: ioOffset += sizeof(hx::Object *);storeType=fsObject;  break;
-         case etVoid:
-         case etNull:
+         case fsBool: ioOffset += sizeof(int); break;
+         case fsByte: ioOffset += sizeof(int); break;
+         case fsInt: ioOffset += sizeof(int); break;
+         case fsFloat: ioOffset += sizeof(Float); break;
+         case fsString: ioOffset += sizeof(String); break;
+         case fsObject: ioOffset += sizeof(hx::Object *); break;
+         case fsUnknown:
```
tobil4sk: Is there a way we could use `sTypeSize` here?

MeguminBOT: I'll check in a bit!
Thanks for feedback

MeguminBOT: Yeah that works, changed it.

test/cppia/cases/TestCommon.hx
```diff
@@ -59,6 +59,59 @@ class TestCommon extends Test {
...
 
+    @:depends(testStatus)
+    function testBoolMemberStorage() {
+        final cls = Type.resolveClass('ClientBoolField');
+
+        if (Assert.notNull(cls, 'Unable to resolve ClientBoolField')) {
+            final obj = Type.createInstance(cls, []);
+
+            Assert.equals('true', Std.string(Reflect.field(obj, 'flag')), 'Member Bool did not read back as a boolean');
+            Assert.equals('true', Std.string(Reflect.field(cls, 'staticFlag')), 'Static Bool did not read back as a boolean');
```
tobil4sk, line 70: The tests in this file seem a bit excessive and probably add more confusion than clarity. Most of the checks are already passing anyway so they aren't really related to the fix. I think these two checks would be sufficient:

```haxe
Assert.same(Type.ValueType.TBool, Type.typeof(Reflect.field(obj, 'flag')), 'Member Bool did not read back as a boolean');
Assert.same(Type.ValueType.TBool, Type.typeof(Reflect.field(cls, 'staticFlag')), 'Static Bool did not read back as a boolean');
```

tobil4sk (approved):
Looks good to me now, the new code is cleaner and more consistent with the static implementation:
https://github.com/HaxeFoundation/hxcpp/blob/10f73c52cb4c82e72df20d012216b1592eff8d28/src/hx/cppia/CppiaVars.cpp#L99

PR #1374, opened 2026-08-20
[windows] file_write: require a console, not just a character device

dimensionscape:
_isatty is true for any character device, so a write to NUL took the WriteConsoleW path. WriteConsoleW fails on a non-console handle and the file_error after it throws from inside a GC-free zone, so the process dies instead of reporting anything:

    sys.io.File.write("NUL").writeString("x")
    prog.exe >NUL

GetConsoleMode only succeeds on a real console, and __hxcpp_print already uses it. _isatty stays as the first check so files and pipes take the same fwrite path they did before.

Came in with 1544ff5e (#1307), which fixed utf console output. That part is unchanged here.


src/hx/libs/std/File.cpp
```diff
@@ -149,7 +149,10 @@ int _hx_std_file_write( Dynamic handle, Array<unsigned char> s, int p, int n )
 
    hx::AutoGCFreeZone zone;
 #ifdef HX_WINDOWS
-   if (_isatty(_fileno(f->io))) {
+   // _isatty is true for ANY character device - NUL, a serial port, a printer - and none of
+   // those accept WriteConsoleW. Only a real console has a console mode.
+   DWORD console_mode;
+   if (_isatty(_fileno(f->io)) && GetConsoleMode((HANDLE)_get_osfhandle(_fileno(f->io)), &console_mode)) {
```
tobil4sk, line 155: Should we keep the `_isatty` check here? can we just rely on `GetConsoleMode`?

PR #1385, opened 2026-09-02
Fix Conversion Warnings with Conservative Marking

Aidan63:
After a comment in https://github.com/HaxeFoundation/hxcpp/pull/1339 this is the first of probably a few PRs where I'm going through the IMMIX GC with all of clangs conversion warnings set to errors and fixing them up.
In this case I've gone through the conservative marking function and what it calls to make them compile even if you enable all of clangs warnings as errors. There are a few things going on here.

- Curly bracket initialisation is now used, this forbids narrowing / lossy conversion which will silently happen other wise.
- `uintptr_t` is used instead of `size_t` for pointer manipulation. As of C++11 this type is the standard sanctioned way for pointer editing.
- C++ style casts are used instead of C-style.

For reference below is the clang pragma I used to enable all the warnings as errors, I put these just inside the `MarkConservative` function with a pragma pop at the end of the function to limit them to just that function and what it calls.

```
#pragma clang diagnostic push
#pragma clang diagnostic error "-Wconversion"
#pragma clang diagnostic error "-Wsign-conversion"
#pragma clang diagnostic error "-Wimplicit"
#pragma clang diagnostic error "-Wimplicit-int-conversion"
#pragma clang diagnostic error "-Wall"
#pragma clang diagnostic ignored "-Wunused-variable"
#pragma clang diagnostic ignored "-Wunused-but-set-variable"
```

### DIFF
--- a/src/hx/gc/Immix.cpp
+++ b/src/hx/gc/Immix.cpp
@@ -5622,7 +5622,7 @@ void MarkConservative(int *inBottom, int *inTop,hx::MarkContext *__inCtx)
    uintptr_t lastPin{};
    #ifdef HX_WATCH
    uintptr_t lastWatch{};
-   bool isWatch{false};
+   bool isWatch{};
    #endif
 
    #ifdef HXCPP_GC_GENERATIONAL
--- a/src/hx/gc/Immix.cpp
+++ b/src/hx/gc/Immix.cpp
@@ -100,11 +100,9 @@ static void *sgObject_root = 0;
 // With virtual inheritance, stack pointers can point to the middle of an object
 #ifdef _MSC_VER
 // MSVC optimizes by taking the address of an initernal data member
-static int sgCheckInternalOffset = sizeof(void *)*2;
-static int sgCheckInternalOffsetRows = 1;
+static uintptr_t sgCheckInternalOffset = sizeof(void *)*2;
 #else
-static int sgCheckInternalOffset = 0;
-static int sgCheckInternalOffsetRows = 0;
+static uintptr_t sgCheckInternalOffset = 0;
 #endif
 
 int gInAlloc = false;
@@ -168,13 +166,13 @@ static size_t sgMaximumFreeSpace  = 1024*1024*1024;
 
 
 #ifdef HX_WATCH
-void *hxWatchList[] = {
-  (void *)0x0000000100000200,
-  (void *)0
+uintptr_t hxWatchList[] = {
+  uintptr_t{ 0x0000000100000200 },
+  uintptr_t{ 0 }
 };
-bool hxInWatchList(void *watch)
+bool hxInWatchList(uintptr_t watch)
 {
-   for(void **t = hxWatchList; *t; t++)
+   for(uintptr_t* t = hxWatchList; *t; t++)
       if (*t==watch)
          return true;
    return false;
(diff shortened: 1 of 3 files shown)


src/hx/gc/Immix.cpp
```diff
@@ -5626,74 +5618,75 @@ void MarkConservative(int *inBottom, int *inTop,hx::MarkContext *__inCtx)
...
 
-   void *prev = 0;
-   void *lastPin = 0;
+   uintptr_t prev{};
+   uintptr_t lastPin{};
    #ifdef HX_WATCH
-   void *lastWatch = 0;
-   bool isWatch = false;
+   uintptr_t lastWatch{};
+   bool isWatch{false};
```
tobil4sk: This seems excessive for bools, is it intended?

src/hx/gc/Immix.cpp
```diff
@@ -5703,47 +5696,46 @@ void MarkConservative(int *inBottom, int *inTop,hx::MarkContext *__inCtx)
...
+                     int y = info->GetEnclosingAllocType(pos-sizeof(int),&potentialObject,allowPrevious);
                      #ifdef HXCPP_GC_NURSERY
-                     void *nptr;
-                     int z = info->GetEnclosingNurseryType(pos-sizeof(int),&nptr);
+                     int z = info->GetEnclosingNurseryType(pos-sizeof(int),nullptr);
                      #else
                      int z = 0;
                      #endif
-                     printf("but got alloc type=%d, enclosing=%d nurs=%d o=%d\n",x,y,z,sgCheckInternalOffset);
+                     printf("but got alloc type=%d, enclosing=%d nurs=%d o=%d\n", x, y, z, static_cast<int>(sgCheckInternalOffset));
```
tobil4sk: Here we probably want to change the format specifier, rather than casting to int

src/hx/gc/Immix.cpp
```diff
@@ -5626,74 +5618,75 @@ void MarkConservative(int *inBottom, int *inTop,hx::MarkContext *__inCtx)
...
 
-      if (vptr && !((size_t)vptr & validObjectMask) && vptr!=prev && vptr!=lastPin)
+      if (potentialObject && !(potentialObject & validObjectMask) && potentialObject != prev && potentialObject != lastPin)
       {
 
          #ifdef PROFILE_COLLECT
          hx::localCount++;
          #endif
-         MemType mem = sGlobalAlloc->GetMemType(vptr);
+         MemType mem{ sGlobalAlloc->GetMemType(potentialObject) };
```
tobil4sk, line 5654: Here the braces also seem unnecessary, since it is MemType -> MemType

tobil4sk (commented):
Seems like a good change. I've also got some WIP local fixes to help get rid of some unrelated warnings in other files, I'm hoping we can eventually have cleaner code and enable warnings more globally.

src/hx/gc/Immix.cpp
Aidan63: I agree but I also don't like mixing and matching the initialisation syntaxes, so I decided to go with this one for all of them.

src/hx/gc/Immix.cpp
Aidan63: Got rid of the false since thats what bools default to.

src/hx/gc/Immix.cpp
Aidan63: I did that because I wasn't sure how uintptr_t worked with printf, turns out you're supposed to use some magic PRIdPTR macro.

src/hx/gc/Immix.cpp
tobil4sk: I understand the need for this syntax when converting between types initialisation, but using it in universally makes the code less natural and less readable imo. However if it's just limited to this marking function then I guess that's fine.

Aidan63: I think ideally I'd rather use auto for all this stuff, e.g.

```
auto pos = size_t{ potentialObject & IMMIX_BLOCK_BASE_MARK };
auto mem = sGlobalAlloc->GetMemType(potentialObject)
```

I think that's consistent and I can align the variable names which I like, but I can't remember if this is C++11 or 14. Stuff around bracket initialisation changed in 14.

PR #1388, opened 2026-09-04
size_t for all alloc functions

Aidan63:
Like my previous PR, but for the alloc functions. All these functions now compile with those clang pragma and `size_t` is used in place of `int`.
One thing I did notice here is that very, very, big allocs will not have a correct size in the object header. 32bits are used in the GC header for a large objects size, so if you allocate an object created than `std::numeric_limits<unsigned int>::max()` it will get an invalid size (line 3300). I don't know if anyone has tried to allocate an object this large, but maybe we should add a check and raise a hxcpp critical error in these cases.

### DIFF
--- a/src/hx/gc/Immix.cpp
+++ b/src/hx/gc/Immix.cpp
@@ -2860,19 +2860,27 @@ bool IsConstAlloc(const void *inData)
    return header[-1] & HX_GC_CONST_ALLOC_BIT;
 }
 
-void *InternalCreateConstBuffer(const void *inData,int inSize,bool inAddStringHash)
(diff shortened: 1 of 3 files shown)


src/hx/gc/Immix.cpp
```diff
@@ -3233,18 +3232,18 @@ class GlobalAllocator
...
          mLargeAllocSpace = inSize<<1;
 
-      unsigned int *result = 0;
+      unsigned int* result{};
       #ifndef HXCPP_SINGLE_THREADED_APP
-      bool do_lock = true;
+      bool do_lock{ true };
       #else
-      bool do_lock = false;
+      bool do_lock{};
```
tobil4sk: I think an explicit false here makes more sense if this value will have an effect

PR #1393, opened 2026-09-09
size_t for gc globals

Aidan63:
This updates those global `sg` variables to be `size_t`. I think looking at switching QuickVec to being `size_t` based will be the thing to tackle next as there's lots of questionable implicit conversions going on with `size` and the likes.

### DIFF
--- a/src/hx/gc/Immix.cpp
+++ b/src/hx/gc/Immix.cpp
@@ -97,7 +97,6 @@ void DebuggerTrap()
 
 static bool sgAllocInit = 0;
 static bool sgInternalEnable = true;
-static void *sgObject_root = 0;
 // With virtual inheritance, stack pointers can point to the middle of an object
 #ifdef _MSC_VER
 // MSVC optimizes by taking the address of an initernal data member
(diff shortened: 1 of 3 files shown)


src/hx/gc/Immix.cpp
```diff
@@ -97,7 +97,6 @@ void DebuggerTrap()
 
 static bool sgAllocInit = 0;
 static bool sgInternalEnable = true;
-static void *sgObject_root = 0;
```
tobil4sk, line 100: Just to verify this removal, it seems this variable has been unused since: f7d27a491ce2c6ec61c06647d927871f0e864991

src/hx/gc/Immix.cpp
```diff
@@ -6556,9 +6555,6 @@ void InitAlloc()
    sgFinalizers = new FinalizerList();
    sFinalizerLock = new std::mutex();
    sGCRootLock = new std::mutex();
-   hx::Object tmp;
-   void **stack = *(void ***)(&tmp);
-   sgObject_root = stack[0];
 
    //GCLOG("__root pointer %p\n", sgObject_root);
```
tobil4sk: We can get rid of this commented out log now, since `sgObject_root` is gone

PR #1398, opened 2026-09-13
Overhaul Register Capture

Aidan63:
I noticed some inconsistencies in terms of how registers are captured and what is captured, so this is an attempt to unify them.

- Windows will always use `RtlCaptureContext` where as previously only 64bit x86 would use it. I have not tested the Arm64 path as I don't have access to an Arm Windows device, maybe there's a VM image somewhere.
- All Windows targets will now capture the floating point registers, previously only 64bit x86 would do this. This is possible due to always using the `RtlCaptureContext` function which gives us access to the FXSAVE structure and NEON registers.
- The `_fxsave` intrinsic is used for non Windows x86 to capture the floating point registers, previously the floating point registers were never captured.
- All general purpose x86 registers are captured, previously the `RegisterCapture::Capture` function was used which contained four arguments to try and make sure the compiler couldn't pass arguments into functions as registers.
- Where possible we capture the stack pointer and use that as the "bottom of the stack" instead of getting the address of a local variable.

Windows is still the only platform in which the Arm NEON registers are captured, I haven't found a nice way like _fxsave to capture all the extended state for Arm.

One issue is that 32bit x86 Windows seems to consistently crash in release mode, if I compile in debug mode it seems to run fine...


Aidan63:
Welp, I'm stumped with this x86 windows crash. I can re-create it in release mode with HXCPP_DEBUG_LINK to get some debugging details, It's occurring in the weak hash tests after a collection. The stack pointer is definitely being set and appears to be correct, and the registers have values in them so that seems to be correct, plus x64 windows has no problem...
I've also tried setting x86 windows to use both the setjmp and inline assembly versions and the same problem occurs.

PR #1399, opened 2026-09-13
[ci] Add linux arm64 build

tobil4sk:
Hxcpp has had support for linux arm for some time, and haxe is already running linux arm hxcpp tests.

For some reason, the cffi test currently fails on arm64 linux:
```
Called from TestCffi::boot src/TestCffi.hx line 17
Called from cpp.Lib::load cpp/Lib.hx line 38
Error : Could not load module prime@isBool__1
```
I'll try to emulate locally to figure it out.


tobil4sk:
The failure is caused by a missing arm64 case in `__hxcpp_get_bin_dir()`. There is an open pr to fix it: #1355.

MRK2009:
You mentioned my pr? Noice
Hope my pr gets merged before yours.

